PR #112, opened 2014-12-16
Added most important pieces of metadata for publication

tripu:
This fixes what's more important about #106.
See [Returning metadata about the document](https://github.com/w3c/specberus/blob/tripu/issue-106-metadata/README.md#returning-metadata-about-the-document) for more information about the way metadata is returned.

PR #125, opened 2014-12-25
Fix some of the comments described in #112

astorije:


README.md
```diff
@@ -27,7 +27,7 @@ $ nodejs app.js [PORT]
 
 Meaning of positional parameters:
 
-1. `PORT`: where Specberus will be listening for HTTP connections.
```
tripu: These two extra spaces at the end of the line weren't a typo. They're [a line break](http://spec.commonmark.org/0.15/#hard-line-breaks), so that default values appear neatly indented below the description of their argument.

But never mind, it's a silly thing :) It doesn't block your PR.

Issue #130, opened 2015-01-09
Deliverers cannot be identified based on extracted data from the document


astorije:
At the moment, Specberus extracts from the documents the name of the WG as it is displayed, and the underlying link, which is the homepage.

Unfortunately, none of these are enough to properly identify a Working Group without heuristics:
- The displayed name can change from one document to another, like *[Web Applications WG](http://www.w3.org/TR/2015/REC-IndexedDB-20150108/)*, *[Web Applications Working Group](http://www.w3.org/TR/2014/WD-DOM-Level-3-Events-20140925/)*, *[Web Applications (WebApps) Working Group](http://www.w3.org/TR/2013/WD-appmanifest-20131217/)*, ...
- The homepage can match to multiple Working Groups in W3C database

I have no good solution to suggest for this, and I don't think a solution. One way (that I dislike) could be to add the ID of the group (or any other unique identifier) to the document similarly to the editors.
Another one (that I like) would be to make the homepage unique. We currently have only 100 link duplicates in the corresponding table, 8 duplicates for homepages that contain `w3.org`.

tripu:
I think the best approach would be to add more metadata to specs, in this case in the form of a new `data-deliverer-id` attribute. It makes perfect sense to expect specs to be machine-readable, if we want them to be ultimately *machine-publishable*.

I suspected this would be the conundrum as soon as we starting adding *some* metadata to specs, and I'm glad that things are pushing us towards more semantic specs.

That being said, I await your comments, and a decision from the group, before tackling this one.

deniak:
+1 on @tripu 's comment. If we are asking the editors to add their id in the document, adding the WG's id shouldn't be too difficult and gives us a good way to identify the deliverers

astorije:
-1 on the WG ID suggestion, as said on my previous message. I agree we need all possible metadata, but the ID makes things obscure and confusing. Do we have another piece of information to identify the group that we can make use of?
I won't fight over the ID suggestion though, we don't have time for this.

I would like to have feedback from @plehegar, @darobin and @dontcallmedom however.

dontcallmedom:
The current checker relies on the WG homepage — I'm not sure why this needs to change now; we can look to a more reliable system overtime, but I don't think that should be a blocking factor for specberus in the short term.

astorije:
[...]

PR #131, opened 2015-01-13
Deliverers' ID attribute recognised by Specberus, and spat out as part of metadata

tripu:
We haven't reached consensus in #130. But @deniak and I favoured reading a new attribute `data-deliverer-id` on documents, and it's the quickest solution for the issue. So here is that solution.

For more info about how those IDs are being returned now, see [the updated documentation](https://github.com/w3c/specberus/blob/tripu/issue-130-deliverers-metadata/README.md#emitting-metadata-about-the-document).

Also, notice that the new attribute **is not required** by Specberus, ie not finding it on the document will *not* throw an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ following fields:
 
 * `url`: URL of the content to check. One of `url`, `source`, `file`, or `document` must be
   specified and if several are they will be used in this order.
-* `source`: A string with the content to check.
+* `source`: A `String` with the content to check.
 * `file`: A file system path to the content to check.
 * `document`: A DOM `Document` object to be checked.
 * `profile`: A profile object which defines the validation. Required. See below.
@@ -85,13 +85,13 @@ These properties are now returned when found:
 * `title`: The (possible) title of the document.
 * `process`: The process rules, **as they appear on the text of the document**, eg `'14 October 2005'`.
 * `deliverers`: The deliverer(s) responsible for the document (WGs, TFs, etc); an `Array` of `Object`s, each one with these properties:
-  * `homepage`: URL of the grop's home page.
+  * `homepage`: URL of the group's home page.
   * `name`: name of the group, exactly as it is found in the hyperlink on the document.
+* `delivererIDs` ID(s) of the deliverer(s); an `Array` of `Number`s.
 * `thisVersion`: URL of this version of the document.
 * `previousVersion`: URL of the previous version of the document (the last one, if multiple are shown).
 * `latestVersion`: URL of the latest version of the document.
-* `editorIDs`: ID(s) of the editor(s) responsible for the document (an `Array` of `String`s, necessarily matching the pattern `/\d+/`).
-* `delivererID` ID of the group, (a `String`, necessarily matching the pattern `/\d+/`).
+* `editorIDs`: ID(s) of the editor(s) responsible for the document; an `Array` of `Number`s.
 
 As an example, validating [`http://www.w3.org/TR/2014/REC-exi-profile-20140909/`](http://www.w3.org/TR/2014/REC-exi-profile-20140909/) (REC)
 emits these pairs of metadata:
@@ -156,8 +156,8 @@ the following metadata will be found:
 
 Profiles are simple objects that support the following API:
 
-* name: A string being the name of this profile.
-* rules: An array of rule objects which are checked in this profile.
+* name: A `String` being the name of this profile.
+* rules: An `Array` of rule objects which are checked in this profile.
 
 A profile is basically a configuration of what to check. You can load a specific profile from under
 `lib/profiles` or create your own.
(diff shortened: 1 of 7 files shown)


astorije:
@deniak, @tripu and I discussed this on the phone today. We will use @dontcallmedom's suggestion to extract the WG ID from the PP link, to get to the point faster, but some of us express concerns about this solution. @plh had ideas and we will most likely refactor after v1.

@tripu, I will review when you are done, no rush :)

tripu:
@astorije: change done; please review.

README.md
```diff
@@ -84,11 +84,14 @@ These properties are now returned when found:
...
 * `process`: The process rules, **as they appear on the text of the document**, eg `'14 October 2005'`.
-* `deliverers`: The deliverer(s) responsible for the document (WGs, TFs, etc).
+* `deliverers`: The deliverer(s) responsible for the document (WGs, TFs, etc); an `Array` of `Object`s, each one with these properties:
+  * `homepage`: URL of the grop's home page.
+  * `name`: name of the group, exactly as it is found in the hyperlink on the document.
 * `thisVersion`: URL of this version of the document.
 * `previousVersion`: URL of the previous version of the document (the last one, if multiple are shown).
 * `latestVersion`: URL of the latest version of the document.
-* `editorIDs`: ID(s) of the editor(s) responsible for the document, necessarily matching the pattern `/\d+/`.
+* `editorIDs`: ID(s) of the editor(s) responsible for the document (an `Array` of `String`s, necessarily matching the pattern `/\d+/`).
```
astorije: Why do you need to specify that this is an `Array` of `String`s? Isn't it possible to send an array of integers instead?

README.md
```diff
@@ -84,11 +84,14 @@ These properties are now returned when found:
...
+* `delivererID` ID of the group, (a `String`, necessarily matching the pattern `/\d+/`).
```
astorije: Why isn't this part of the `deliverers`? Something like:
- `deliverers`
  - `id`
  - `homepage`
  - `name`

Also, there can be multiple deliverers, so there will be multiple IDs detected.

lib/rules/heuristic/group.js
```diff
@@ -4,16 +4,28 @@
...
 exports.check = function (sr, done) {
-    var patterns = /.+ interest group$|.+ community group$|.+ working group$/i;
-    var candidates = [];
-    var item;
-    var i;
+    var patterns = /.+ interest group$|.+ community group$|.+ working group$/i
+    ,   candidates = []
+    ,   candidate
+    ,   item
+    ,   i;
```
astorije: This is such an ugly, error-prone syntax! (I am not blaming you but the very first guy who used it :) )

FYI, I plan to switch them back all to the traditional `var [...];` one. This horror has to stop and the world needs a leader :D

lib/rules/sotd/pp.js
```diff
@@ -64,17 +70,25 @@ exports.check = function (sr, done) {
...
-        if (/^http:\/\/www\.w3\.org\/2004\/01\/pp-impl\/\d+\/status$/.test(href) &&
-            text === "public list of any patent disclosures" &&
-            $a.attr("rel") === "disclosure") {
-            foundPublicList = true;
-            return;
+        if (/^http:\/\/www\.w3\.org\/2004\/01\/pp-impl\/\d+\/status(#.*)?$/.test(href) &&
+            text === "public list of any patent disclosures") {
+            ids = DELIVERER_ID_REGEX.exec(href);
+            console.log (href + ' - ' + DELIVERER_ID_REGEX);
+            console.dir(ids);
```
astorije: Aren't these 2 lines remains from your development tests? If not, how useful is it in general? This is not sarcasm this time, I am genuinely wondering :)

astorije:
@tripu, review done. Mostly, a bit puzzled about the `delivererID` scalar param versus the `deliverers` collection param.

tripu:
@astorije, [I've committed a few changes](https://github.com/w3c/specberus/commit/09c4cafd06f001d89a729f61e9efcf3e2c862c8e), in response to your comments.
Details:

* `delivererIDs` and `editorIDs` are now `Array`s of `Number`s (not `String`s like before).
* Now, multiple `delivererIDs` are returned (it's a vector instead of an scalar). Those IDs can't be part of the `Object`s inside `deliverers` because those two pieces of information are extracted in different ways, from different parts of the document. There's no (consistent) way to associated IDs to names and homepages: there could be a different number of IDs and homepages, and even if the number is the same, how do we know which is which&hellip;?
* Removed `console.log()` and `console.dir()` statements (oops! thanks!).

Additional fixes:

* `README.md`: updated with latest changes to metadata fields, fixed hierarchy of document profiles, fixed typo, improved styling.
* Profile `TR`: fixed a bug (rule `sotd/pp` was being checked twice).

astorije:
Looks good, I just suggest a fix in #135.

PR #132, opened 2015-01-13
Show a hyperlink to old pubrules

tripu:


astorije:
This seems straightforward and I don't have specific comment about it. If someone complains about how things are displayed, I think this will need to be addressed later. I merge.

PR #138, opened 2015-01-28
Retrieve the hierarchy of tracks and profiles in the UI from a new JSON file

### LINKED ISSUE #51: The hierarchy of available profiles is hard-coded in the UI
Values and labels in the drop-down menu should be populated somehow from the set of profiles under `lib/profiles/` &mdash; we may need to add some metadata to those profiles for that.

This would simplify maintenance of profiles and rules, and avoid mistakes resulting from the redundancy that we have now. Admittedly, it would also introduce new complexity.

tripu:
Fixes #51. (Necessary/convenient to tackle the addition of `status` to the metadata returned by Specberus.)

* Tracks and profiles are now loaded using a `GET` request for a new local JSON file, and the `<select>` input gets populated from that data.
* Fixed a bug: the value of CG notes was `FPWG-NOTE` by mistake.
* The *prompt* message of the `<select>` input is now non-selectable.

### DIFF
--- a/public/data/profiles.json
+++ b/public/data/profiles.json
@@ -1,27 +1,27 @@
 {"tracks": [
     {"name": "Recommendations",
      "profiles": [
-         {"id": "WD", "name": "Working draft"},
-         {"id": "FPWD", "name": "First public working draft"},
-         {"id": "FPLC", "name": "First public and last call working draft"},
-         {"id": "LC", "name": "Last call working draft"},
-         {"id": "CR", "name": "Candidate recommendation"},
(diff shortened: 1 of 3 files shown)


public/data/profiles.json
```diff
@@ -0,0 +1,27 @@
+{"tracks": [
+    {"name": "Recommendations",
+     "profiles": [
+         {"id": "WD", "name": "Working draft"},
```
deniak: Better to capitalize (first letter upper case) the name of each profile.

Issue #106, opened 2014-11-21
Extract metadata for publication


astorije:
Here are the needed pieces of data:

* **TR status** (WD implied for V1; refuse FPWD and LCWD)
* **Date** (metadata checker: Check that pub date is "today"; problem with TZ: everybody must use ET)
* **Title**
* **Process Rules** (2014 process by default)
* **This version**
* **Latest version**
* **Previous versions** (first one, multiple ones?)
* **Deliverers** (multiple deliverers, not for v1; metadata checker: check that data from Specberus and from registration match)
* **Editors** (ID or username? **Done: ID**)

astorije:
@tripu, as far as I could see, Specberus doesn't send the editor's draft URL. Could you add it please? Thanks!

deniak:
@astorije @tripu The current pubrules doesn't check anything regarding the editor's draft. The easiest would be to add a rule "just" for the extraction but we are starting to mix different things imho.

astorije:
@deniak, agreed. Do you have an idea on how we could avoid this?
I am personally thinking of a sub-library only here for extraction and both Specberus and Echidna rely on it. Since that I have little experience with Specberus, I would not be surprised to be wrong :) In addition, this sounds like the current pubrules checker architecture, proven to be difficult to maintain over time (but probably for other reasons).
The other issue I see is that it's a much bigger feature and in that case your easy route would do the job meanwhile.

tripu:
Having two separate projects/libraries (one for metadata extraction, another one for proper validation) vs. having Specberus doing both was a design decision that we discussed during F2F. And IIRC, we opted for extending Specberus with data extraction, because the way the document would be traversed (and regex'es applied) would be pretty similar in both cases, but at the same time refactoring that in a separate library would be too much work.

So for the time being, I'll add the editor's draft to the list of metadata being returned (if it's feasible). Would that be OK?

astorije:
Perfectly okay by me, @tripu.

tripu:
@astorije: if you approve and merge #127, shall we close this issue?

Maybe move the remaining pieces of (secondary) metadata to a new issue, with label `v1.5` or `enhancement`&hellip;?

astorije:
I think some other issues might arise and it will be best to discuss about them here, in my opinion. So I would vote for leaving this thread open and, whenever I have a question on the subject, I'll post here.

tripu:
[...]

PR #139, opened 2015-01-28
Return “status” (a “String”) as part of metadata; update documentation accordingly

tripu:
See issue #106.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ These properties are now returned when found:
 * `previousVersion`: URL of the previous version of the document (the last one, if multiple are shown).
 * `latestVersion`: URL of the latest version of the document.
 * `editorIDs`: ID(s) of the editor(s) responsible for the document; an `Array` of `Number`s.
+* `status`: ID (acronym) of the profile detected in the document; a `String`. See file `public/data/profiles.json`.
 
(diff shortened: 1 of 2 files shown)


tripu:
@astorije, in case you're wondering why Specberus returns the profile acronym [as such](https://github.com/w3c/specberus/blob/master/public/data/profiles.json), instead of returning only *{WD, PER, CR, FPLC, REC}*, @deniak and I seemed that it makes sense to return this information with the granularity we have &mdash; no need to group profiles. Some of them can be treated the same later in the endpoint, if that's required&hellip;

astorije:
That's a valid argument indeed. We'll limit to `WD` in Echidna and will see what this affects in the future, but for now it makes no difference there.

astorije:
Oh and since I'm here... We are not updating the version number at the moment, so when doing `npm install` on echidna, npm thinks there was no modification even if it's not true.

This time I solved this by doing explicit `npm install specberus` but I think `npm install -f` works too. I couldn't find a nicer way to have the `npm install` by default getting the latest changes from the repo (since echidna defines the dependency as a Github repository). I wonder if there is a way to improve this, either on specberus's side (like a special `0.0.0` version or something else defined in its `package.json`) or on echidna's side (when defining the dependency).

@deniak, @plehegar, @darobin, any idea? (and also, a warning, don't forget to update echidna...)

PR #151, opened 2015-02-23
Fix #145: Supports old and new forms, as well as transition

### LINKED ISSUE #145: New copyright form
New rules are in effect (we are at the beginning of a one-month transition period) during which "All Rights Reserved" is no longer part of the copyright string. The old pubrules suggests a copyright string without that phrase (but then fails to validate without it), so that people are already using it (ReSpec will stop generating it with the next release). Specberus needs to update to match.

astorije:


astorije:
@deniak, @tripu, @darobin, who wants to be assigned to this one? :-)

lib/rules/headers/copyright.js
```diff
@@ -9,15 +9,23 @@ exports.check = function (sr, done) {
...
-        ,   evil = "All Rights Reserved\\."
-        ,   good = "Some Rights Reserved: this document is dual-licensed, CC-BY and W3C Document License\\."
-        ,   endRex = " W3C liability, trademark,? and document use rules apply\\.$"
-        ,   rex = new RegExp(startRex + (hatesFreedom ? evil : good) + endRex)
+        ,   startRex = "^Copyright © (?:(?:199\\d|20[01]\\d)-)?" + year + " .*?W3C® \\(MIT, ERCIM, Keio, Beihang\\)"
+        ,   dualLicensed = /CC-BY/.test(text);
+
+        // See https://lists.w3.org/Archives/Public/spec-prod/2015JanMar/0001.html
+        if (sr.getDocumentDate() < new Date(2015, 01, 07)) // Before 06 Feb 2015
+            var w3cOnlyStatement = ", All Rights Reserved";
```
darobin: You're declaring this variable three times (and the three declarations get hoisted to the function's top). Just declare it as part of the top block and then simply assign to it in the conditional.

lib/rules/headers/copyright.js
```diff
@@ -9,15 +9,23 @@ exports.check = function (sr, done) {
     if ($c.length) {
         var year = (sr.getDocumentDate() || new Date).getFullYear()
         ,   text = sr.norm($c.text())
-        ,   startRex = "^Copyright © (?:(?:199\\d|20[01]\\d)-)?" + year + " .*?W3C® \\(MIT, ERCIM, Keio, Beihang\\), "
-        ,   hatesFreedom = /All Rights Reserved/.test(text)
-        ,   evil = "All Rights Reserved\\."
-        ,   good = "Some Rights Reserved: this document is dual-licensed, CC-BY and W3C Document License\\."
-        ,   endRex = " W3C liability, trademark,? and document use rules apply\\.$"
```
darobin: You take all the fun out :)

darobin:
With the minor `var` thing fixed this LGTM.

lib/rules/headers/copyright.js
astorije: FUN IST VERBOTEN

lib/rules/headers/copyright.js
astorije: Fixed in cd48bc2

astorije:
@darobin, thanks! I wasn't sure regarding the dates to compare to (aka will people be upset when their document fails on March 8...).

darobin:
I guess we'll see when this happens. It has been announced and there's been a transition period. The tools are updated (at least ReSpec is, @tabatkins can confirm for Bikeshed). We should be okay!

mikewest:
Bikeshed is updated as of https://github.com/tabatkins/bikeshed/commit/d80da506293c08b3effb85cfa950be8b7332ade1.

darobin:
Thanks @mikewest! Well, @astorije, nothing stands in the way of merging :)

PR #153, opened 2015-02-26
Make an Echidna-specific profile for WDs

astorije:
This adds an Echidna-specific profile, checking that the document being checked:

- has `data-editor-id` attributes for the editors
- is set at today's date

It comes with a few tests. I couldn't find a way to automate tests that the current date passes the checker. If you have suggestions, I'm all ears, otherwise I tested manually and it passes...

On top of the usual review, let me know if names and file organization is fine as I am not used to Specberus (@darobin++ for the architecture, btw!) yet.

Once you are okay with the inner changes, I will add the UI part so that people can use Specberus with Echidna profile(s). Makes sense?

### DIFF
--- a/lib/profiles/WD-Echidna.js
+++ b/lib/profiles/WD-Echidna.js
@@ -2,14 +2,7 @@
 // Working Draft profile
 
 exports.name = "WD-Echidna";
-exports.config = {
-    status:             "WD"
-,   longStatus:         "Working Draft"
-,   previousVersion:    true
-,   styleSheet:         "W3C-WD"
-,   stabilityWarning:   true
-,   recTrackStatus:     true
-};
+exports.config = require("./WD").config;
 
 var base = require("./base");
 exports.rules = base.insertAfter(
--- a/lib/profiles/WD-Echidna.js
+++ b/lib/profiles/WD-Echidna.js
@@ -1,0 +1,22 @@
+
+// Working Draft profile
+
+exports.name = "WD-Echidna";
+exports.config = {
+    status:             "WD"
+,   longStatus:         "Working Draft"
+,   previousVersion:    true
+,   styleSheet:         "W3C-WD"
+,   stabilityWarning:   true
+,   recTrackStatus:     true
+};
+
+var wd = require("./WD");
+exports.rules = wd.insertAfter(
+        require("./TR").rules
+    ,   "sotd.status"
+    ,   [
+            require("../echidna/editor-ids.js")
+        ]
+);
+
--- a/lib/profiles/WD-Echidna.js
+++ b/lib/profiles/WD-Echidna.js
@@ -17,6 +17,7 @@ exports.rules = wd.insertAfter(
     ,   "sotd.status"
     ,   [
             require("../echidna/editor-ids.js")
+        ,   require("../echidna/todays-date.js")
         ]
 );
 
--- a/lib/profiles/WD-Echidna.js
+++ b/lib/profiles/WD-Echidna.js
@@ -11,13 +11,13 @@ exports.config = {
 ,   recTrackStatus:     true
 };
 
-var wd = require("./WD");
-exports.rules = wd.insertAfter(
-        require("./TR").rules
+var base = require("./base");
+exports.rules = base.insertAfter(
+        require("./WD").rules
     ,   "sotd.status"
     ,   [
-            require("../echidna/editor-ids.js")
-        ,   require("../echidna/todays-date.js")
+            require("../rules/echidna/editor-ids.js")
+        ,   require("../rules/echidna/todays-date.js")
         ]
 );
 
--- a/app.js
+++ b/app.js
@@ -21,6 +21,7 @@ var express = require("express")
 ;
 ("FPWD FPLC WD LC CR PR PER REC RSCND " +
 "CG-NOTE FPIG-NOTE IG-NOTE FPWG-NOTE WG-NOTE " +
+"WD-Echidna " +
 "MEM-SUBM TEAM-SUBM").split(" ")
          .forEach(function (p) {
              profiles[p] = require("./lib/profiles/" + p);
--- a/lib/l10n.js
+++ b/lib/l10n.js
@@ -179,6 +179,9 @@ for extended information about the publication rules.'
     ,   "heuristic.group.not-found":          "Could not identify a group behind this spec."
 	// heuristic/date-format
     ,   'heuristic.date-format.wrong':          'This date found on a boilerplate section of the document does not have the expected format (<code>DD Month YYYY</code>, with an optional leading zero in the day): <em>&ldquo;${text}&rdquo;</em>.'
+    // Echidna
+    ,   'echidna.editor-ids.no-editor-id': 'Editor IDs must be set for each editor, in the <code>data-editor-id</code> attribute.'
+    ,   'echidna.todays-date.wrong-date': 'The publication date of this document must be set to today.'
     }
 };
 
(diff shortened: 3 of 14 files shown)


deniak:
@astorije we should be able to use that profile from the web interface. It'll be easier to check the documents instead of going through echidna.
The profile used in echidna (L41)[https://github.com/w3c/echidna/blob/master/functions.js#L41] should also be updated but I guess you are waiting for this PR to be merged before.

lib/profiles/WD-Echidna.js
```diff
@@ -0,0 +1,23 @@
...
+exports.name = "WD-Echidna";
+exports.config = {
+    status:             "WD"
+,   longStatus:         "Working Draft"
+,   previousVersion:    true
+,   styleSheet:         "W3C-WD"
+,   stabilityWarning:   true
+,   recTrackStatus:     true
+};
+
```
darobin: You probably don't need to copy the configuration, either. You can just copy the `config` exported by `WD.js`.

darobin:
@deniak The idea we had when hashing out this approach was that the UI might not list "Echidna" as one option in the drop-down. First, it's a bit cryptic, and second if there are too many options it gets bothersome.

Instead we thought of a checkbox for "Valid for automatic publication (with link)". When checked, that would reduce the list of options to just "WD", and the server would know to pick up the Echidna variant. It's more work but it seems more user-friendly.

lib/profiles/WD-Echidna.js
astorije: I'll look into that, thanks!

astorije:
@deniak, what @darobin said!

> @astorije we should be able to use that profile from the web interface. It'll be easier to check the documents instead of going through echidna.

Yes, I am planning to add that soon. You can either merge this beforehand if you want to quick fix the underlying issue and I'll open another PR for the UI, or you can wait for everything to be done. I'd say first option but it's your call :-)

> The profile used in echidna should also be updated but I guess you are waiting for this PR to be merged before.

Right, I will!

astorije:
@darobin, @deniak, how does b84e3f7 look like regarding UI?

I have one problem though, I lose the state (profiles disabled) when I reload the page. I tried something but got into weirder results. I prefer to have your feedback before, and maybe help if possible...

astorije:
> * You keep using `select#profile` as a selector; it's wasteful because the ID matches anyway. Also, it's better to cache repeated selectors in a variable.

Fixed in ff943c0. Better?

astorije:
> * You keep using `select#profile` as a selector; it's wasteful because the ID matches anyway. Also, it's better to cache repeated selectors in a variable.

Fixed in ff943c0, better?

astorije:
> * You're not maintaining state on reload because you're not getting a change event simply by setting the `checked` property. In turn `disableProfilesIfNeeded` isn't called.

Yeah, I meant when I tried locally. I didn't commit my tests as they were unsuccessful. Should I spend time on this now or can we merge this and have it fixed when you refactor? We kind of really need to merge this, and the UI is becoming a collateral damage...

darobin:
Yes, go ahead and merge!

tripu:
@darobin:
> * What's the ` />` at the end of your `input` element?

@astorije [complained about my ` />` before, too](https://github.com/w3c/echidna/pull/68#discussion_r24122912). I didn't answer then, but let me grab this opportunity to give you my reasons, and explain **why I think always closing all elements is a good practice**. Not a big deal&nbsp;&mdash;&nbsp;I'm ready to concede and let you guys have it your way :¬)

Not being &ldquo;necessary&rdquo; doesn't mean it's &ldquo;wrong&rdquo;; removing those ` />` doesn't &ldquo;fix&rdquo; anything.

Note that the same argument you guys are using (*not needed*, *what for*) should prompt us to rewrite our code to get rid of (for example) &ldquo;unnecessary&rdquo; quotation marks around many attributes, eg
```html
<img height="48" width="72" alt="W3C" src="http://www.w3.org/Icons/w3c_home">
```
would be unnecessary, and should be &ldquo;fixed&rdquo;:
```html
<img height=48 width=72 alt=W3C src="http://www.w3.org/Icons/w3c_home">
```
etc. Not to mention unnecessary semicolons in JS, other syntactic sugar, etc.

For the price of a few extra keystrokes here and there, we get closer to polyglot documents, and contribute to make it a bit easier for both *html* and *XML* parsers. In general, I think it's a good idea to follow the [guidance for XHTML-HTML compatibility](https://wiki.whatwg.org/wiki/HTML_vs._XHTML) when that doesn't cost us a thing. There is even a philosophical point to be made here; a statement, or a precautionary tale for other languages or platforms :¬) I guess I feel [like Bruce](http://html5doctor.com/html-5-xml-xhtml-5/) about this:
> I like the xhtml syntax. It’s how I learned. I’m used to lowercase code, quoted attributes and trailing slashes on elements like br and img. They make me feel nice and comfy.

So, to sum up my opinion: the fact that we gave up to bad markup long ago and accepted it as a necessary evil doesn't imply that we shouldn't aim at robust syntax when there is nothing to lose (and at least a tiny bit to gain) in doing so.

astorije:
I don't want to be seen as trolling here: I am actually very curious at @darobin's answer, as he is nothing less than one of the editors of the [HTML5 spec](http://www.w3.org/TR/html5/). This is IMO a legit question that many HTML developers regularly ask and I would like to know what definite rationale to bring on that.

There are a few things you said @tripu that I think are not true. I could just let @darobin answer but, actually, I'd prefer him to tell me I'm wrong on what I can answer. Although I'll look dumb, I'll learn things along the way :-)

> get rid of (for example) “unnecessary” quotation marks around many attributes

These are actually necessary in some cases, like `alt="W3C is da bomb"`, where the trailing slash on [void elements](http://www.w3.org/TR/html5/syntax.html#void-elements) is never necessary. Apart from that and personal preferences, I don't know if one should prefer with or without quotes.

> Not to mention unnecessary semicolons in JS, other syntactic sugar, etc.

Semicolons are not unnecessary in JS. And actually, it's the opposite: semicolon omission is not a syntactic feature, automatic semicolon insertion is :-) And while you want to know what are the rules for ASI, you really don't want to [mess with it](http://inimino.org/~inimino/blog/javascript_semicolons).

> we get closer to polyglot documents, and contribute to make it a bit easier for both html and XML parsers.

Not long ago, I heard an argument that if HTML5 doesn't look like well-formed XML, it's good because it's not XML, and will never be. From what I understood, the XML-like syntax will be less and less "followed", to make the best of HTML's future possibilities.

> we gave up to bad markup long ago and accepted it as a necessary evil

Sigh...

---

I can't wait to be proven wrong by @darobin or any other HTML5+ expert around here :-)

darobin:
Self-closing slash is different from quotes. I like to use quotes because that way I don't need to think about what the attribute value contains, it can all be regular. There are reasons for JS semicolons too, notably script concatenation problems.

By contrast, self-closing elements does nothing. Indeed, if you were to want to parse your document with an XML parser, it could help. But why would you want to do that? What's your use case? Are you taking steps to ensure that the document really is fully parseable as XML? Are you taking steps to ensure that the same DOM will be produced? Using `CDATA` in scripts and making sure to always include `tbody` in tables? Are you planning to send the content as `application/xhtml+xml`?

If the answer to any of the above is "no", then there really is no point in self-closing. It's a false sense of comfort. The document can't be processed as XML, and in fact for correctness and safety it *should* not.

On the other hand, if the answer to all of the above (and then some) is "yes", I'm going to have to ask "Why?" :)

PR #156, opened 2015-03-02
Update whacko and remove old hacks

### LINKED ISSUE #152: Inline bug reports regarding whacko that do not match whacko's documentation
- [rules/headers/dl](https://github.com/w3c/specberus/blob/c08a08b4022d34a91c4216a421665d69b8f2bc98/lib/rules/headers/dl.js#L29) says:

> `// XXX one of the horrible hacks that can be removed when whacko supports .next properly`

but [whacko seems to support it](https://github.com/cheeriojs/cheerio#nextselector).

---

- [rules/structure/h2](https://github.com/w3c/specberus/blob/8aac15c70444a9742ff7ad99239686899467735a/lib/rules/structure/h2.js) says:

> `// whacko doesn't really do .parents(selector)`

but [whacko says they do](https://github.com/cheeriojs/cheerio#parentsselector).

---

- [lib/rules/style/sheet.js](https://github.com/w3c/specberus/blob/8aac15c70444a9742ff7ad99239686899467735a/lib/rules/style/sheet.js) says:

> `// nextAll() doesn't work in whacko *cries*`

but [whacko doesn't say such thing](https://github.com/cheeriojs/cheerio#nextall).

---

@darobin, I wonder if these warnings only applied with a specific version, or in a specific use case, or similar, and they can be fixed now. If not, did you find GitHub issues regarding these problems that we can link to?
Also, if this happens because of a certain use case, any reason to think other functions in the API are likely to not work as well?

That'd be nice if we can at least document here why these are present if we cannot do anything about that.

astorije:
This fixes #152.

The latest API allows us to get rid of some hacks, which is always nice :-)

### DIFF
--- a/lib/rules/headers/dl.js
+++ b/lib/rules/headers/dl.js
@@ -26,25 +26,6 @@ exports.check = function (sr, done) {
     ,   err = function (str, extra) {
             sr.error(exports.name, str, extra);
         }
-        // XXX one of the horrible hacks that can be removed when whacko supports .next properly
-    ,   getDD = function ($dl, idx) {
-            var pastDT = false
-            ,   dtIdx = -1
-            ,   $dd = null
-            ;
-            $dl.children().each(function () {
-                var $el = sr.$(this);
-                if (pastDT && $el.is("dd")) {
-                    $dd = $el;
-                    return false;
-                }
-                if (!pastDT && $el.is("dt")) {
-                    dtIdx++;
-                    if (dtIdx === idx) pastDT = true;
-                }
-            });
-            return $dd;
-        }
     ;
 
     if (subType === "member") topLevel = "Submission";
@@ -56,7 +37,7 @@ exports.check = function (sr, done) {
                       .replace(":", "")
                       .toLowerCase()
                       .replace("published ", "")
-        ,   $dd = getDD($dl, idx)
+        ,   $dd = $dt.next('dd')
         ,   key = null
         ;
         if (!$dd) return err("no-dd", { dt: txt });
--- a/lib/rules/structure/h2.js
+++ b/lib/rules/structure/h2.js
@@ -3,18 +3,8 @@ exports.name = "structure.h2";
 exports.check = function (sr, done) {
     var h2s = [];
     sr.$("h2").each(function () {
-        var $h2 = sr.$(this)
-        ,   inHead = false
-        ;
-        // whacko doesn't really do .parents(selector)
-        $h2.parents().each(function () {
-            var $el = sr.$(this);
-            if (($el[0].name || $el[0].tagName).toLowerCase() === "div" && $el.hasClass("head")) {
-                inHead = true;
-                return false;
-            }
-        });
-        if (!inHead) h2s.push(sr.norm($h2.text()));
+        var $h2 = sr.$(this);
+        if ($h2.parents("div.head").length === 0) h2s.push(sr.norm($h2.text()));
     });
     if (h2s[0] !== "Abstract") sr.error(this.name, "abstract", { was: h2s[0] });
     if (!/^Status [Oo]f [Tt]his [Dd]ocument$/.test(h2s[1])) sr.error(this.name, "sotd", { was: h2s[1] });
--- a/lib/rules/style/sheet.js
+++ b/lib/rules/style/sheet.js
@@ -13,12 +13,9 @@ exports.check = function (sr, done) {
         ]
     ,   $lnk = sr.$(sels.join(", "))
     ;
-    if ($lnk.length) {
-        // console.log("found", sr.$(sel + " ~ link[rel=stylesheet]"));
-        // nextAll() doesn't work in whacko *cries*
-        if (sr.$(sels.map(function (s) { return s + " ~ link[rel=stylesheet]"; }).join(", ")).length)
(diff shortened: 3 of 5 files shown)


lib/validator.js
```diff
@@ -192,13 +192,13 @@ Specberus.prototype.getSotDSection = function () {
...
+        if ($startH2[0] === this) {
             started = true;
             return;
         }
         if (!started) return;
-        if ($endH2[0] === this[0]) return false;
-        $div.append(this.clone());
+        if ($endH2[0] === this) return false;
+        $div.append($(this).clone());
     });
```
astorije: Updating whacko from 0.13.x to 0.17.x made this piece of art break. I'm not sure why the changes above fixed the stuff, but it did. I couldn't find where this change became needed (along with whacko, a ton of dependencies evolved) so if you are aware of something, please share. I had a hard time fixing this...

In a nutshell, now `this[0]` is `undefined`, voilà.

lib/rules/style/sheet.js
```diff
@@ -13,12 +13,9 @@ exports.check = function (sr, done) {
         ]
     ,   $lnk = sr.$(sels.join(", "))
     ;
-    if ($lnk.length) {
-        // console.log("found", sr.$(sel + " ~ link[rel=stylesheet]"));
-        // nextAll() doesn't work in whacko *cries*
-        if (sr.$(sels.map(function (s) { return s + " ~ link[rel=stylesheet]"; }).join(", ")).length)
-            sr.error(this.name, "last");
+    if (!$lnk.length) sr.error(this.name, "not-found");
+    else if ($lnk.nextAll("link[rel=stylesheet]").length) {
```
astorije: See cheeriojs/cheerio#674

lib/rules/headers/dl.js
```diff
@@ -56,7 +37,7 @@ exports.check = function (sr, done) {
                       .replace(":", "")
                       .toLowerCase()
                       .replace("published ", "")
-        ,   $dd = getDD($dl, idx)
+        ,   $dd = $dt.next('dd')
```
astorije: This one is pretty cool, isn't it? :-)

lib/rules/structure/h2.js
```diff
@@ -3,18 +3,8 @@ exports.name = "structure.h2";
...
-        $h2.parents().each(function () {
-            var $el = sr.$(this);
-            if (($el[0].name || $el[0].tagName).toLowerCase() === "div" && $el.hasClass("head")) {
-                inHead = true;
-                return false;
-            }
-        });
-        if (!inHead) h2s.push(sr.norm($h2.text()));
+        var $h2 = sr.$(this);
+        if ($h2.parents("div.head").length === 0) h2s.push(sr.norm($h2.text()));
```
astorije: There might be a better way to filter out the `h2` elements that are in the `div.head`, in which case I am all ears.

astorije:
I think the failed tests are failed only because of the timeouts, but I'd like to be sure before it's merged (it times out on my laptop too)... Let me know if you have any concerns.

lib/validator.js
darobin: In jQuery, `each` sets `this` to the DOM element (not its jQuery wrapper). Clearly whacko wasn't doing that and I worked around it (I don't recall the issue, but it looks that way). It seems that whacko is now working a lot more like jQuery. This could potentially bite for any other instance of `each`.

astorije: Argh... You want me to do a pass on everything else? We might want to rely on the second argument of the function passed as an argument (`.each(function (index, element) { /* ... */ })`) instead of `this`, I think it would be safer.

lib/rules/headers/dl.js
darobin: :rocket: :guitar:

lib/rules/structure/h2.js
darobin: This looks fine to me.

darobin:
I'm getting 10 failed tests and they don't look like they're from timeouts.

```
  Category sotd
    Rule supersedable
      1) should pass for headers/simple.html
      2) should fail for sotd/supersedable.html
    Rule mailing-list
      3) should pass for headers/simple.html
      4) should fail for sotd/ml-bad.html
    Rule pp
      5) should pass for headers/simple.html
      6) should fail for sotd/pp-bad.html
      7) should pass for sotd/pp-cpp2002.html
    Rule stability
      8) should pass for headers/simple.html
      ✓ should fail for headers/simple.html 
      ✓ should pass for sotd/supersedable.html 
    Rule implementation
      ✓ should pass for sotd/supersedable.html 
      9) should pass for sotd/pp-bad.html
      ✓ should fail for sotd/ml-bad.html 
    Rule ac-review
      10) should pass for sotd/supersedable.html
      ✓ should fail for sotd/pp-bad.html 
```

lib/validator.js
darobin: I did a pass on everything else and it's all fine.

astorije:
Argh... You didn't forget to `npm install`, did you? :P

lib/validator.js
astorije: \o/

darobin:
No, but it made no difference. However `rm -Rf`ing `node_modules` and running install again worked. BRUTALITY FTW!

PR #159, opened 2015-03-10
Bugfix: check that the document date is valid before using it

tripu:
When trying to validate a bogus document, eg `http://foo.bar`, a `TypeError` crashes Specberus: `cannot call method "setHours" of null`. It happens at least when reloading the page after the first attempt, ie when the document is a parameter in the URL.

The error is in `todays-date.js:6`: `sr.getDocumentDate()` isn't a `Date`, so we can't do `.setHours(0, 0, 0, 0)` with it.

### DIFF
--- a/lib/rules/echidna/todays-date.js
+++ b/lib/rules/echidna/todays-date.js
@@ -3,7 +3,11 @@
 exports.name = "echidna.todays-date";
 
 exports.check = function (sr, done) {
-    function getDate(date) { return date.setHours(0, 0, 0, 0); }
+
+    function getDate(date) {
+        if (date) return date.setHours(0, 0, 0, 0);
+        else return null;
+    }
 
     if (getDate(sr.getDocumentDate()) !== getDate(new Date())) {
         sr.error(exports.name, 'wrong-date');


astorije:
When can we have no date set? Might that be when the date is not properly detected? If so, is there a way to not run this test at all?

astorije:
I don't like this solution, among other things because it violates the type safety of the function.

So the problem comes from the fact that in some cases `sr.getDocumentDate()` is not an actual date. At best, we should be able to not run this test if `sr.getDocumentDate()` is incorrect, or to have it necessarily fail. That would be ideal.

If the current architecture doesn't allow us to do my above suggestion, an alternative could be the following:

```js
if (!(sr.getDocumentDate() instanceof Date)) {
         sr.error(exports.name, 'date-not-detected');
}
else if (getDate(sr.getDocumentDate()) !== getDate(new Date())) {
         sr.error(exports.name, 'wrong-date');
}
```

with the according `date-not-detected` string in the error descriptions.

Also, could you squash your previous 3 commits into 1 please?

tripu:
@astorije: I think most of our functions don't have clear contracts, eg about checking/assuming validity of inputs, or delegating parameter sanitisation, etc. I don't feel strongly about that, either. Mine was just a simple solution for a fatal runtime error.

So feel free to commit to this branch if you think it should be improved!

About squashing commits: definitely a good idea, thank you for reminding me. I won't do it here because I don't like to rewrite history that's already pushed to a remote. But I'll keep that in mind next time, before I push commits to open a new PR!

astorije:
@tripu, type safety is not about contracts set in a project :-) Anyway, I'll do the changes tomorrow assign the PR to you.

Regarding rewriting the history, it's never a good idea to do it on a shared/master branch, but always a good idea on a feature/personal branch, before it makes it to master. It's okay to rewrite history if you know that nobody will push in that branch before `pull`ing. I have `rebase`d more than I can remember :-P

tripu:
Thanks, @astorije. I'll review and merge if you want. But feel free to merge with your changes yourself. This doesn't deserve too much time from us :¬)

astorije:
@tripu, see #163. I will merge it as soon as Travis is happy to make Echidna's test suite run directly. Feel free to comment there if something looks wrong to you.

@plehegar told me he will try to do w3c/echidna#100, which is related, by Friday, to make sure I didn't mess anything up :-)

PR #161, opened 2015-03-10
Fix editor ids wrong return type

astorije:


astorije:
Merging myself now since it is breaking tests on Echidna (https://github.com/w3c/echidna/issues/105).

PR #163, opened 2015-03-11
Fix the today's date check

astorije:
This extends #159 by leaving the function type safe, fixing the function as it was mutating the source `Date` object, and adding a check/error if the date cannot be extracted from the document.


tripu:
Tested, and looks OK to me.

Assigning now to @plehegar as this depends on w3c/echidna#100, according to @astorije.

astorije:
Actually, it's the opposite. This ticket must be merged (and version `1.0.1` must be published) so that @plehegar can work on https://github.com/w3c/echidna/issues/100.

I am doing it now so that it doesn't impact @plehegar's progress.

PR #168, opened 2015-03-16
implement insafe instead of safe-url-input-checker module

guibbs:
Hi,

Specberus was using the safe url input checker module (to check if an url is well formed on the server side) which is deprecated.
Instead, I wrote a new Promise based API more efficient: [insafe](http://github.com/w3c/insafe) and implemented it in Specberus.

### DIFF
--- a/app.js
+++ b/app.js
@@ -15,7 +15,7 @@ var express = require("express")
 ,   l10n = require("./lib/l10n")
 ,   util = require("util")
 ,   events = require("events")
-,   urlSafety = require("safe-url-input-checker")
+,   insafe = require("insafe")
 ,   version = require("./package.json").version
 ,   profiles = {}
 ;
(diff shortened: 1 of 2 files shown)


app.js
```diff
@@ -92,14 +92,14 @@ io.sockets.on("connection", function (socket) {
...
-        urlSafety.checkUrlSafety(data.url, function(err, res) {
-            if(!res) {
-                socket.emit("exception", {message: "error while resolving " + data.url + " Check the spelling of the host, the protocol (http, https) and ensure that the page is accessible from the public Internet."});
-            }
-            else {
+        insafe.check({
+            url: data.url,
+            statusCodesAccepted: ["301"]
+        }).then(function(res){
+            if(res.status == true) {
```
astorije: In general, please use `===` instead of `==` to avoid type coercion.

But in that case, `if (res.status)` is enough.

astorije:
:+1:

PR #169, opened 2015-03-16
Fix #164: Check the validation tests conditionally if the test suite does not run within Travis CI

### LINKED ISSUE #164: Travis builds keep failing
The validation tests (like this [css validator timeout](https://travis-ci.org/w3c/specberus/jobs/53991933) or this [html validator timeout](https://travis-ci.org/w3c/specberus/jobs/53882940)) very often fails due to timeouts. This is something we don't really want in general, but so far we could ignore the errors and pretend things work fine. It is not great because I remember myself not checking a build, being sure it was a timeout error, where it wasn't...

The problem is, #158 brought automatic publication on npm after a new tag is pushed **when a build is successful**. That could be problematic, as if a build fails, the automatic publication will not kick in.

Any idea on how to beat these timeouts?

astorije:


astorije:
@darobin, I am assigning this PR to you as you mentioned the idea on #164. Does that match what you were thinking of?

astorije:
*Please merge me*

tripu:
Yes, we are often getting false positives from Travis CI. They aren't a big deal, but they are annoying sometimes.

So if this fixes that, I'd merge now, @astorije!

darobin:
SHIIIIP IT!

Issue #155, opened 2015-03-02
Subheading using h2


darobin:
The spec templates for W3C make use of `h2` for a subheading (with the status and date). According to HTML, that's wrong though: http://www.w3.org/TR/html5/common-idioms.html#sub-head

Needless to say, the HTML specification would like to adhere to this. We've been using `p.subline` instead, with a matching style sheet being:

```
p.subline {color:#005A9C; font: 140% sans-serif;margin: 0.75em 0;}
```

I would like to suggest making this the default across the board; and accepted by Specberus.

plehegar:
That seems fine but:
1- we need to update both specberuc and the /TR style sheets
2- we should keep allowing h2 for a transition period or until the old pubrules get retired

Can you start a thread on this on spec-prod to announce the intent?

astorije:
Transition periods are *the best*, can we haz more?

plehegar:
you cannot change the rule under the feet of the editors and expect everyone to follow them in the next hour. Anolis, respec, and bikeshed will also need to be updated. We're dealing with an ecosystem here, not with a team of individuals.

darobin:
@astorije Well, they'll always be annoying but I think the solution you came up with last time is good. At least for this one we can simply make use of the same.

In fact, we could make it a core functionality. Something like:

```
st.transitoryCheck({
    startDate:   "2015-03-15"
,   endDate:    "2015-04-15"
,   old:             function () {}
,   between     function () {}
,   new:             function () {}
});
```
It would automatically call the right option based on the document's date. We could even add functionality such that if the end date is more than one year ago, it produces a warning to remove the code or some such.

astorije:
I knooow, and I'm sure @darobin knew that too [when we discussed about it](https://github.com/w3c/specberus/issues/145#issuecomment-73728843). But still, until I have a idea to deal with transitions in a smarter way, [it leads to awkward code](https://github.com/w3c/specberus/blob/cd48bc266505e5e4fd921e932a5dc909032d1e4b/lib/rules/headers/copyright.js#L20) IMO.

astorije:
Ouh, I'd like something like that @darobin! I'll think of your suggestion :-)

darobin:
That's always the solution @astorije: when there's awkward code, hide it in a dependency!

astorije:
@darobin, now that #170 is merged, I wanted to solve this but I'm a bit confused: what exactly do you want me to change on Specberus?

[...]

PR #170, opened 2015-03-16
Add a transition period feature

astorije:
@darobin, does that match [what you meant](https://github.com/w3c/specberus/issues/155)?

I think I'd like to see this feature also handle changes that do not need a transition periods, if it ever happens. Do you see how we could adapt this wording to deal with that as well?

Also, do you think we need to add any check on the input, what's stored in `options` (`from` and `to` are `Date` objects, `do*` are `function` values, ...)?

Once this is final and merged, and you have reached a conclusion on spec-prod, I will use this for #155.


darobin:
I think that this is good as is, thanks! I doubt that checking the input will help that much.

Issue #29, opened 2014-06-20
Check against inconsistencies in URLs of this/previous/latest versions


astorije:
Two related checks:

1. [x] Check against inconsistency when *previous version URI* ≠ *redirected latest version URI*.
2. [x] *This version URI* ≠ *previous version URI*.

**About the first one:**

- **OK** when Previous Version of the document to be published == the document behind the Latest Version
- **FAILURE** otherwise.

Edge cases:

- Once the document is published, this rule cannot return OK anymore.
- PER: after publication, the Latest Version must not change (point to the REC rather than the published PER).

Hopefully I will come up with a PR someday, but I want to agree on a decision before anyway.

**About no. 2:**

Echidna [has published (wrongly) already a spec with *this version URI* = *previous version URI*](http://www.w3.org/TR/2015/WD-appmanifest-20150317/).

plehegar:
Feel free to assign an issue to me in echidna for this case to be tested once this issue is resolved.

astorije:
I do not recognize my *hand*-writing in what is written in the original issue so I'm assuming its content was changed and refactored. I decline property of this issue then :-)

Due to GitHub being flexible here and its lack of history for issues, let's please not edit each other's comments besides small typos.

tripu:
@astorije Yes, I edited the title & description of the issue to make it clearer&nbsp;&mdash;or so I thought&mdash;, and to accommodate that new bug that is closely related&hellip; I was careful and kept almost all your wording. But I'm sorry if that confused things. Because there's no history of those changes as you say, maybe we should add comments (even if they change the meaning of the issue altogether), yes&hellip;

marcoscaceres:
Hey, not sure where to file this, but I the manifest spec was affected by this bug: 

http://www.w3.org/TR/2015/WD-appmanifest-20150317/

Is it possible to fix that for the spec above? I've had someone report it to me.

tripu:
Yes, @marcoscaceres&nbsp;&mdash;&nbsp;that spec was precisely the one that brought this bug to @deniak's attention.

I see that there have been already a few versions published after that one, so I'm assuming you are using the system OK again now&hellip;?

About correcting the &ldquo;previous version&rdquo; link for that particular version, I guess that is the right thing to do here, and it'll take just a minute. But I'm pinging @astorije (ping!) here first, to make sure it's OK if I do that myself now, manually.

marcoscaceres:
[...]

PR #173, opened 2015-04-01
Check against inconsistencies in URLs of this/previous/latest versions

tripu:
I'm comparing the *contents* (the actual documents) of "previous" vs. "latest"
&hellip;but just the *URLs* of "this" vs. "previous".

I understand that is what will guarantee that we're catching all those tricky situations explained in issue #29&hellip;? I have also extended the tests to try to catch these cases.

(@plehegar, it'd be great if you want to extend the Echidna test suite to cover these cases, once/if this is merged :¬)

### DIFF
--- a/lib/rules/headers/dl.js
+++ b/lib/rules/headers/dl.js
@@ -8,6 +8,9 @@
 //  h2 date and this version date must match
 //  dt for editor or author
 
+var Request = require('request')
+,   Promise = require('promise');
+
 exports.name = "headers.dl";
 
 exports.check = function (sr, done) {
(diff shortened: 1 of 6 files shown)


lib/rules/headers/dl.js
```diff
@@ -133,6 +156,21 @@ exports.check = function (sr, done) {
...
+      }
+    }, function() {
+      err('cant-retrieve');
+    });
+
+    if (thisURI.toLowerCase() === previousURI.toLowerCase()) {
+      err('same-this-and-previous');
+    }
+
     done();
```
astorije: Shouldn't `done()` be inside the `then()` function, since it's asynchronous?

astorije:
On top of my inline comment, I'd like a :+1: from @darobin before giving mine :-)

tripu:
@astorije, you were right, thanks! I think I just fixed this issue, and now `done()` is invoked if/when promises succeed or fail, too.

@darobin, assigning to you as @astorije suggested.

astorije:
@darobin, you have my :+1:, feel free to merge yourself or let me do it!

tripu:
Merged myself after checking with @deniak (and having had @astorije's blessing too)&nbsp;&mdash;&nbsp;this one's been sitting here for too long, and I'd like to have it in master to deploy a new release this week.

PR #180, opened 2015-04-09
Two minor fixes

tripu:
* Deleted redundant, deprecated selectors file.
* Removed hyperlink to `/` for the title in the header (eg, when in `https://labs.w3.org/pubrules/` it links to `https://labs.w3.org/`, which doesn't exist).


public/index.html
```diff
@@ -20,7 +20,7 @@
         <div class="row">
           <div class="col-md-8 navbar-brand">
             <a href="http://www.w3.org/"><img class="logo" src="img/logo-w3c.svg" alt="W3C logo"></a>
-            <a href="/"><img src="img/logo.svg" alt="Logo for the Pubrules Checker">Pubrules Checker</a>
+            <img src="img/logo.svg" alt="Logo for the Pubrules Checker">Pubrules Checker
```
astorije: I agree with you that linking to `/` was wrong, but it is quite common practice to link to the main page (and in that case, helpful to "clear" a report).

Anyway, I'll send a PR with my proposal, feel free to accept or refuse :-)

PR #183, opened 2015-04-10
Betterify the hyperlink to "reset" the app (follow-up #181)

tripu:
The hyperlink added in PR #181 is broken *unless the app happens to be deployed at `[host]/pubrules/`*. Fixing that.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -20,7 +20,7 @@
         <div class="row">
           <div class="col-md-8 navbar-brand">
             <a href="http://www.w3.org/"><img class="logo" src="img/logo-w3c.svg" alt="W3C logo"></a>
-            <a href="/pubrules/"><img src="img/logo.svg" alt="Logo for the Pubrules Checker">Pubrules Checker</a>
+            <a href="#" onclick="location = location.href.replace(/\?.*$/g, '')"><img src="img/logo.svg" alt="Logo for the Pubrules Checker">Pubrules Checker</a>
           </div>
           <div class="contribute col-md-4 text-right">
             <a id="infoLink" href="#" onclick="$('#about').show(); $('#infoLink').hide();">Important information about this tool</a>


astorije:
@tripu, sorry #181 got merged without your approval. I was surprised to see @darobin merging rogue :D

Regarding this PR, as much as I hate doing so, I think I need to :-1: here. Although I am not a JS-hater (that would be pointless considering the rest of the page is full JS), I don't feel comfortable having an inline JS string for a simple link.

I think it should stay that way, nevermind if it's broken if you install it some place else. What does @darobin think about this?

darobin:
Merging rogue? I wasn't aware we had rules, I probably didn't get that memo :)

astorije:
Anarchist!

astorije:
@darobin what do you think of this PR?

darobin:
Can we back out to the goal here? It looks like the intent is to effectively reset the parameters. Wouldn't it be better to add a reset button to the form?

Also: who cares if the link only works when deployed there? I would suggest that anyone developing locally should also use `/pubrules/` locally, otherwise you might make a mistake with an include or something.

astorije:
> Can we back out to the goal here? It looks like the intent is to effectively reset the parameters. Wouldn't it be better to add a reset button to the form?

Nah, not really. I agree a reset button is needed, but that's not the goal here IMO.
The goal is just to provide a link to the "entry page" behind the navbar logo, which I see as a standard practice on pretty much all the websites today. I surprise myself being frustrated when such link doesn't exist, because that became a mechanism for me to go back to the index page.

> Also: who cares if the link only works when deployed there? I would suggest that anyone developing locally should also use `/pubrules/` locally, otherwise you might make a mistake with an include or something.

Right, not really a problem if it only works when deployed.
I also :+1: your suggestion to use `/pubrules/` locally. Is there something that needs to be changed on `app.js` (or else) to set this as a base?

tripu:
OK. I'm outnumbered here; closing this one :¬)

Further talk about forcing deployment at `/pubrules/`, or adding a reset button, in separate issues/PRs.

PR #188, opened 2015-04-24
send an error if a whitespace doesn't follow the long status in h2

deniak:


darobin:
Actually there is ample precedent for allowing this: https://github.com/tabatkins/bikeshed/pull/400/files

deniak:
I guess it doesn't hurt to allow the comma although [pubrules](http://www.w3.org/2005/07/pubrules?uimode=filter&uri=) explicit says
```
The document's status and date MUST be in an h2 element as follows (see also date syntax):

<h2>W3C Working Draft 14 August 2015</h2>
```
Since we allowed the comma in the past, I updated the PR to accept it.

PR #185, opened 2015-04-15
Quickfix #178: filter out URLs from paragraphs when checking for well-formed dates

astorije:
This doesn't fix everything, in particular it doesn't solve[@darobin's last comment on the subject](https://github.com/w3c/specberus/issues/178#issuecomment-91503953), but this provides a quick fix for #178 until something better shows up, featuring [this weird trick that will make your URLs disappear](http://stackoverflow.com/questions/23571013/how-to-remove-url-from-a-string-completely-in-javascript).

PR #189, opened 2015-04-28
Improve #185

tripu:
Suggesting tiny improvements over #185:

* Exclude URLs in uppercase, too (eg `HTTP://...`).
* Add to sample doc more varied examples of excluded dates in URLs.
* Fix coding style (move comma to the beginning of next line).


astorije:
I would personally hit someone who publishes a document with `HTTPS://example.com/yyyy/mm/dd/` in the first place, but agreed that at the moment it's an easy way to by-pass the check.

:+1: Good job, tagging as bug and merging!

PR #192, opened 2015-05-13
README.md: move the badges after the title

deniak:
The description on [npmjs](https://www.npmjs.com/~w3c) is messed up because of the badges in `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
+# Specberus
+
 [![npm version](https://img.shields.io/npm/v/specberus.svg)](https://npmjs.org/package/specberus)
 [![License](https://img.shields.io/npm/l/specberus.svg)](LICENSE)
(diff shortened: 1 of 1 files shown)


tripu:
Well spotted, @deniak.

astorije:
Good catch indeed, but I'm afraid I have to :-1: this PR as I don't think it's the way to go: I would prefer to keep the badges on the first line (for convention), and we don't have control on where the description used by npm stops.

Instead, I'd prefer to set the `description` tag of `package.json`. I did that in #195. I assign both PRs to @tripu to be the referee and decide which PR you want to merge and which one to discard :-)

tripu:
I think the solution suggested in #195 is better.

deniak:
Indeed, it's better to not rely on the `README.md`. Thanks @astorije

PR #198, opened 2015-05-19
Upgrade socket.io to a production release; upgrade also Bootstrap and jQuery

tripu:
* Upgrade socket.io to latest stable (`1.3.5`).
* Upgrade Bootstrap to latest stable (`3.3.4`).
* Upgrade jQuery to latest stable (`2.1.4`).

socket.io reached, and surpassed, `1.0.0`. Specberus was still using a much older version.

--

One caveat that I had to solve when I upgraded:

In the old version (`0.9.16`), [socket.io swallowed `'error'` events](https://github.com/Automattic/socket.io-client/blob/0.9.16/lib/socket.js#L452). Now (`1.3.5`), [`'error'` is propagated, re-emited](https://github.com/Automattic/socket.io/blob/1.3.5/lib/socket.js#L392).

The problem is, [Node.js treats that specific event in a special way](https://nodejs.org/api/events.html#events_class_events_eventemitter):

> *&ldquo;When an EventEmitter instance experiences an error, the typical action is to emit an 'error' event. Error events are treated as a special case in node. If there is no listener for it, then the default action is to print a stack trace and exit the program.&rdquo;*

So the first time that there is a validation error, the Node.js app crashed.

What I did is [change the ID of our `'error'` events to `'err'` when that is passed from the server to the client](https://github.com/w3c/specberus/commit/ce5ea7e0287143afe9b9bd585be8fc29b6e2c511). Besides, that is consistent with the event `'err'` that is being passed already from `lib/validator.js` to `app.js`.


deniak:
@tripu can't we get rid of [`public/js/socket.io.js`](https://github.com/w3c/specberus/blob/tripu/miscellanea/public/js/socket.io.js) ?

tripu:
@deniak, yes&nbsp;&mdash;&nbsp;much in the same way that we could get rid of `public/js/jquery-2.1.4.js` or `public/css/bootstrap.css`, for example. We could remove all those unobfuscated, unminified sources, and the source maps.

Having them in the repo makes it easier to switch from production to development, eg to debug CSS or JS issues. Since they're here already, I'd keep them. But I'm OK if you guys prefer to remove them.

astorije:
:+1: Congrats on updating socket.io!

I'd also :+1: on removing the non minified versions as well as this `.map` file. In practice, I never needed the non-minified versions for debugging on any of my projects, but maybe you had too, so I'll leave the decision to you.

darobin:
I don't use them either. I don't have a strong opinion either way tbh.

PR #199, opened 2015-05-19
Use labs.w3.org's new specific favicon

tripu:


public/index.html
```diff
@@ -11,6 +11,7 @@
     <title>Pubrules Checker</title>
     <link rel="stylesheet" type="text/css" href="css/bootstrap.min.css">
     <link rel="stylesheet" type="text/css" href="css/specberus.css">
+    <link rel="shortcut icon" type="image/x-icon" href="img/favicon.ico" />
```
astorije: Argh, a surviving trailing slash!

darobin: You also don't need `shortcut` :)

tripu:
Thanks, @astorije &amp; @darobin!
Addressed in #200.

PR #202, opened 2015-05-29
Kill kill kill the coverage stat

astorije:
So the coverage seems to have dropped by a lot, but it's actually the *real* coverage now. Similar to [that discussion on Echidna](https://github.com/w3c/echidna/pull/156).


astorije:
I'm done, @tripu!

tripu:
I see you deleted minified files, Flash fallbacks for socket.io, etc, @astorije.
OK to that, since I was the only one who preferred to keep them.

PR #209, opened 2015-07-09
Check dependencies against nodesecurity.io DB as part of CI

tripu:
Run [`nsp audit-package`](https://github.com/nodesecurity/nsp#nsp-audit-package) during the CI cycle.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ node_js:
   - "0.10"
 after_script:
   - npm run coveralls
+  - npm run nsp
 notifications:
   email:
     - denis@w3.org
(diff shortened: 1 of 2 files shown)


tripu:
@deniak, this is ready to merge. The build is (rightly) broken; nsp is doing its job. I'm updating our dependencies in a separate PR to fix that in a moment.

(In spite of [what nsp recommends](https://github.com/nodesecurity/nsp#installation), I prefer to install nsp only locally. If nsp returns OK silently, it's because of [this issue in nsp](https://github.com/nodesecurity/nsp/issues/74#issuecomment-120215841). A solution may be to `rm -rf node_modules/nsp` and simply `npm install` again.)

Issue #208, opened 2015-07-07
Allow exceptions on a per specification basis


deniak:
The CSS WG intends to publish specifications containing example using values (not checked yet by the validator) defined in the specification.

Specberus should be able send warnings instead of errors if an exception has been granted for a specification (CSS copyright, copyrights, IPP, etc). See also #18 and #19.

plehegar:
Here is what I intend to do here: add a json spec config file where, for a given short name, it contains an array of exceptions per rule. If the regex contained in the exception matches the error message generated by the rule, then transform that error into a warning before continuing.

plehegar:
https://github.com/w3c/specberus/pull/261

tripu:
Closed via #261.

PR #212, opened 2015-07-14
Initial support for exceptions

plehegar:
This allows exceptions per spec, eg

```json
  "hr-time-2": [
    {
      "key": "error",
      "name": "validation.css"
    },
    {
      "key": "error",
      "extra":
      {
        "message": "Attribute for not allowed on element a"
      },
      "name": "validation.html"
    }
  ]
```

Addresses #208


tripu:
(Description edited only to get JSON formatting properly.)

lib/exceptions.js
```diff
@@ -0,0 +1,54 @@
+/*jshint es5:true */
```
tripu: `'use strict';` here?
(Or is it already enabled by default in the versions of Node.js we're using?)

app.js
```diff
@@ -12,12 +12,14 @@ var express = require("express")
...
 ,   io = require("socket.io").listen(server)
 ,   Specberus = new require("./lib/validator").Specberus
+,   Exceptions = require("./lib/exceptions").Exceptions
 ,   l10n = require("./lib/l10n")
 ,   util = require("util")
 ,   events = require("events")
 ,   insafe = require("insafe")
 ,   version = require("./package.json").version
 ,   profiles = {}
+,   shortnameRegex = /http:\/\/www.w3.org\/TR\/([-A-Za-z0-9_]+)\//g;
```
tripu: A more robust regex (IMHO):
`/^https?:\/\/(www\.)?w3c?\.org/tr\/([\-A-Za-z0-9_]+)\//gi`

tripu:
Thank you, @plehegar! See the couple of minor comments inline about your changes.

Apart from that, I have these suggestions (most important at the top):

* iiuc, this new logic (have *errors* be only *warnings* when they're granted by an exception) works only when using Specberus stand-alone, ie its web interface. For all this to work also when used as an npm package, eg from Echidna, these changes should be in the class `Specberus`: filtering these errors in [`Specberus#error()`](https://github.com/w3c/specberus/blob/master/lib/validator.js#L91), etc. I think `app.js` would remain unchanged.
* As it is now, errors are only &ldquo;filtered&rdquo; if we know the shortname. But that happens only [when the rule `headers/dl` returns metadata `latestVersion`](https://github.com/w3c/specberus/blob/master/lib/rules/headers/dl.js#L125). There are quite a few rules being applied [before that one](https://github.com/w3c/specberus/blob/master/lib/profiles/base.js#L32), for all profiles. I think exceptions for those would not work.
* Which brings me to my next suggestion: I would prefer to have this logic isolated in a new rule, one that would be applied at the very beginning, for every profile (as the first element in [the base `rules` array](https://github.com/w3c/specberus/blob/master/lib/profiles/base.js#L26)). That way, we make sure the shortname will be known by the time all the other rules are applied. As this isn't really a *rule*, I'd put it inside `rules/heuristic/`.
* Finally, I'm OK with your JSON definition (with *keys* and *names*), but I fear that matching UI messages too (*&ldquo;Attribute for not allowed on element a&rdquo;*) is going to be fragile and error-prone (because comparing strings of natural language verbatim always is). If someone simply fixes a typo in a validator, or in a L10n file of Specberus, or if the description of an error is rewritten to make it clearer for users&hellip; those are merely stylistic changes that should not cause a bunch of specs to suddenly stop validating, I think. Not sure what is the solution here, though. Ideas appreciated!

plehegar:
@tripu re matching UI messages: it gives us the needed granularity imho. we don't want to systematically wipe out all error messages from a validator for example. But the messages should be general enough, ie matching just an element name or a property value for example.

tripu:
@plehegar: agreed. Just to clarify, I wasn't saying we should be *less granular*. What I meant is that maybe we should match IDs of errors that are invariable, short, case-insensitive (instead of trying to match verbatim a long string of *natural language*, which is subjected to a lot of editing over time).

In your example above: if there were an error, or a typo, in that string (`Attribute for not allowed on element a`), or simply if the maintainers of the HTML checker decided to slightly rewrite error descriptions to make them more helpful, or to include links&hellip; the exception granted above would suddenly stop working. I know that doesn't happen often, but still it's a sign that parts would be coupled in a fragile way. Even if the maintainers of Specberus received notice in advance, this kind of changes would require manual, error-prone copy-and-pasting of long strings between different projects (HTML checker, CSS checker, Specberus). Finally, this relying on chunks of text in English would prevent the whole thing from being L10n-ed at some point in the future, or at least make it much more complex.

I also realise that emitting those new error IDs, on top of the existing messages intended for humans, isn't a quick change.

@deniak, @darobin: what do you think? Should we compromise for now, the way that @plehegar is proposing? I only fear that we'll find in the near future some of the problems described above.

What about my other comments, @plehegar? The others were more important, I think.

darobin:
@sideshowbarker Isn't it possible to get machine-readable errors out of the validator? It would be useful for this case.

sideshowbarker:
> @sideshowbarker Isn't it possible to get machine-readable errors out of the validator? It would be useful for this case.

Sure. A simple way is just a GET with `&out=json` included in the query params:
https://validator.w3.org/nu/?doc=https%3A%2F%2Fwww.w3.org&out=json

There’s also an API for doing it through a POST.

Some relevant documentation:

https://github.com/validator/validator/wiki/Service:-HTTP-interface
https://github.com/validator/validator/wiki/Service:-Input:-POST-body
https://github.com/validator/validator/wiki/Service:-Input:-GET
https://github.com/validator/validator/wiki/Service:-Common-parameters
https://github.com/validator/validator/wiki/Output:-JSON

darobin:
@sideshowbarker Sorry, that's not what I meant (Specberus already uses the JSON output). What I meant is a machine-readable way of understanding which error is which. For example, the first error in the link you give has "Bad value “handheld, all” for attribute “media” on element “link”: Deprecated media type “handheld”. For guidance, see the Media Types section in the current Media Queries specification." as the error. That's likely to change, it's human-oriented. I would be looking for something like:

```
{
   type:  "bad-value"
,  el:      "link"
,  attr:    "media"
}
```

(Or whatever works.) This would make it easier to target specific errors to ignore.

sideshowbarker:
> I would be looking for something like:
> ```
> {
>   type:  "bad-value"
> ,  el:    "link"
> ,  attr:  "media"
> }
> ```

Ah. Yeah, we don’t have that and we’re not likely to have it any time in the foreseeable future.

(The current architecture of the code doesn’t lend itself at all to implementing it, and while the benefit of having it is obvious, it’s not obvious to me that the benefit would be worth the pretty significant cost of adding it—both in terms of the implementation costs and also the additional code complexity it’d create.)

darobin:
On 05/08/2015 13:39 , Michael[tm] Smith wrote:
> Ah. Yeah, we don’t have that and we’re not likely to have it any time in
> the foreseeable future.
>
> (The current architecture of the code doesn’t lend itself at all to
> implementing it, and while the benefit of having it is obvious, it’s not
> obvious to me that the benefit would be worth the pretty significant
> cost of adding it—both in terms of the implementation costs and also the
> additional code complexity it’d create.)

Stupid question: is there a locale file for those messages?

-- 
Robin Berjon - http://berjon.com/ - @robinberjon

sideshowbarker:
> Stupid question: is there a locale file for those messages?

Nope. And there’s currently no localization mechanism of any kind for them.

If you’re curious, more details: Some of the messages the checker emits come from 3rd-party libraries, such as https://github.com/smola/galimatias, and we essentially just grab those and pass them on as-is. So the mechanism accepts any arbitrary messages those libraries might generate; we don’t, e.g., maintain a catalog of what messages we expect to see from them, or anything like that.

Anyway the problem with not being able to localize the messages nor to expose structure to allow other consuming apps to richer processing of them is a recognized deficiency in the design, which Henri discussed in his thesis: https://hsivonen.fi/thesis/html5-conformance-checker.xhtml#localizability

The only practical solution for this deficiency under the current architecture would be the “reverse templating” approach that Henri describes in his thesis.

tripu:
Thank you for those explanations and links, @sideshowbarker.

So it is as I suspected&hellip; I'd say these deficiencies in the HTML checker block support for exceptions in Specberus/Echidna, for [all the reasons explained](https://github.com/w3c/specberus/pull/212#issuecomment-123181373) (and a few more: I didn't know the checker also emits UI messages coming from third-party components).

We could try to match verbatim chunks of ~~natural language~~ *one particular natural language* as this PR suggests, but I think that would be short-sighted and costly in the long term. Such change should be temporary and experimental, and groups relying on those exceptions should be aware of that, IMHO. Thus I advise against it.

Maybe we have to take a step back (to #208) and see if there is a better way to get the CSS WG on board with automatic publication?

@deniak, @plehegar: what do you think?

darobin:
Note that the CSS WG is only asking for exceptions for the CSS checker. This might happen to be more manageable in this case than the HTML checker?

tripu:
Who would be the best person to answer these same questions about the CSS checker?

darobin:
Paging @ylafon!

ylafon:
You can use the json output of the css validator, where you have:
for an error
1/ the source, an URL
2/ the line number the error appears
3/ the context
4/ the type of error (foo.bar)
5/ the error message

For warnings (I guess you don't need to care much about those), you don't have the context, and you have a warning level instead (pretty unuseful).

So you have plenty of thing to create a good match for filtering things.

tripu:
@ylafon points me to [the L10n file of the CSS checker](https://github.com/w3c/css-validator/blob/master/org/w3c/css/util/Messages.properties.en) as a reference.
Also, here there is [one example of JSON output from the checker](https://jigsaw.w3.org/css-validator/validator?uri=http%3A%2F%2Fwww.ibm.com%2F&profile=css3&usermedium=all&warning=1&vextwarning=&lang=en&output=json).

@plehegar, what do you think? Could you rewrite your PR to &ldquo;mute&rdquo; certain errors from the CSS checker instead (if that's [what the CSS WG needs](https://github.com/w3c/specberus/issues/208))?

plehegar:
I'll resubmit a new PR to only mute CSS ones instead.

tripu:
OK to close this PR then, @plehegar?

PR #214, opened 2015-07-15
Upgrade Superagent; fix nsp error

tripu:
The history of commits is a bit messy; better to examine only the final diff.

### DIFF
--- a/lib/rules/validation/css.js
+++ b/lib/rules/validation/css.js
@@ -21,13 +21,18 @@ exports.check = function (sr, done) {
     var req
     ,   ua = "Specberus/" + sr.version + " Node/" + process.version + " by sexy Robin"
     ;
-    req = sua.get(service)
-             .set("User-Agent", ua);
     if (sr.url) {
+        req = sua.get(service)
+                 .set("User-Agent", ua);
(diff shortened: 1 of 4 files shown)


deniak:
If the `source` is provided, it should be a `POST` instead of a `GET`

tripu:
@deniak, if I leave `POST` for some of the request (as it used to be), I get the errors [that I described](https://lists.w3.org/Archives/Team/w3t-sys/2015JulAug/0110.html). I thought `GET` was OK too in all cases, so I did that change.

I just changed it back here locally to `POST` when we use `sr.source`, and I get 4 errors:

```
Category validation
  Rule css
    1) should pass for validation/simple.html
    2) should pass for validation/css.html
    3) should fail for validation/bad-css.html
  Rule html
    ✓ should pass for validation/simple.html (1341ms)
    4) should fail for validation/invalid.html
```

It's most likely an error of mine when I updated code to adapt to the new version of Superagent. Still, where is the documentation about this? [`https://jigsaw.w3.org/css-validator/api.html`](https://jigsaw.w3.org/css-validator/api.html)?

PR #216, opened 2015-07-22
Only deploy to npm when the 0.12 build ends

astorije:
This doesn't fix [your recent issue](https://github.com/travis-ci/travis-ci/issues/4197#issuecomment-123582302) but at least it makes sure that the deployment is not run twice each time, the second one always failing.


tripu:
Thanks, @astorije!

PR #225, opened 2015-08-04
Same day publication

### LINKED ISSUE #190: previousVersion link check is incorrect if same day publication
specberus establishes if the previous version is correct by checking the bodies:
[[
        else if (bodies[0] !== bodies[1]) {
]]
lib/rules/headers/dl.js

with bodies[0] coming from latest version
and bodies[1] coming from previousVersion link in the requested publication

However, if the document is published twice the same day through echidna, we need to use for bodies[0] a previous day version, not the latest version...

deniak:
If the previous and latest versions don't match, check if it's a same day publication.

Fix #190


tripu:
In this branch, `npm test` gave me inconsistent errors related to HTML/CSS validation. They seem to be related to checkers timing out, and I'm connected with an unstable link right now, so I'm assuming tests pass OK.

PR #228, opened 2015-08-12
Support for joint publication

### LINKED ISSUE #186: Support for multi-groups publications
Right now, specberus doesn't support publications delivered by more than one groups; that means such publications (which are relatively frequent nowadays) can't be published via Echidna.

An example of such a publication would be http://www.w3.org/TR/2015/WD-mediacapture-depth-20150129/

deniak:
Fix #186 

see https://lists.w3.org/Archives/Team/team-legal/2015Jun/thread.html#msg20


lib/l10n-en_GB.js
```diff
@@ -103,14 +103,16 @@ exports.messages = {
...
+,   "sotd.pp.no-sotd":           "No SotD section."
+,   "sotd.pp.no-pp":             "Required patent policy paragraph not found."
+,   "sotd.pp.no-feb5":           "No link to 2004 Feb 5 PP."
+,   "sotd.pp.no-disclosures":    "No link to public list of disclosures."
+,   "sotd.pp.no-claims":         "No link to definition of essential claims."
+,   "sotd.pp.no-section6":       "No link to section 6."
+,   "sotd.pp.no-jan24":          "No link to 2002 Jan 24 PP."
+,   "sotd.pp.no-pp-transition":  "No link to PP transition procedure."
+,   "sotd.pp.joint-publication": "Joint publication detected."
+,   "sotd.pp.expected-pp":       "Expected text related to patent policy requirements: <blockquote><p>${expected}</p></blockquote>"
```
tripu: Having `blockquote`, isn't `p` unnecessary here? (same in `l10n-es_ES.js`)

public/css/specberus.css
```diff
@@ -82,3 +82,7 @@ td.results a {
     color: #000000;
 }
 
+blockquote {
+  font-size: small;
+  background-color: #ddd;
+}
```
tripu: This background colour looks a little bit too dark for me to read comfortably; I suggest something like `#f0f0f0`.
Also, I'd add `border: 0;` here to get rid of a little left margin that is visible in this block in a different colour.

tripu:
@deniak, this LGTM!
Only a couple minor suggestions; see my comments.
Please decide what to do with those yourself, then merge.

deniak:
@tripu thanks. I updated the markup and the style as you suggested.

PR #229, opened 2015-08-12
Fix boolean conversion causing `"false"` to be converted to `true`

### LINKED ISSUE #206: Loading Specberus with parameters doesn't always produce the same results
**Case 1**

1. Load [Specberus](https://labs.w3.org/pubrules/)
2. Enter in the form:
  * `http://www.w3.org/TR/appmanifest/`
  * Check &ldquo;Check that the document is valid for automatic publication with Echidna&rdquo;
  * WD
3. Result: 2 errors (it's OK, they're related to wrong dates; this is a published WD)

**Case 2**

1. Load the (supposedly) equivalent URL: [`https://labs.w3.org/pubrules/?url=http%3A%2F%2Fwww.w3.org%2FTR%2Fappmanifest%2F&profile=WD-Echidna&validation=simple-validation&noRecTrack=false&informativeOnly=false&echidnaReady=true&patentPolicy=pp2004&processDocument=2014`](https://labs.w3.org/pubrules/?url=http%3A%2F%2Fwww.w3.org%2FTR%2Fappmanifest%2F&profile=WD-Echidna&validation=simple-validation&noRecTrack=false&informativeOnly=false&echidnaReady=true&patentPolicy=pp2004&processDocument=2014)
2. Result: 3 errors. The new one is in the section &ldquo;status of this document&rdquo;:  
**&ldquo;sotd.pp error Required patent policy paragraph not found.&rdquo;**

--

See also #194.

deniak:
Loading specberus directly with query params wasn't checking the document with the right parameters.
The bug comes from a string to boolean conversion where `"false"` was converted to `true`.

Fix #206

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -54,8 +54,8 @@ Specberus.prototype.validate = function (options) {
     self.config.validation = options.validation;
     self.config.htmlValidator = options.htmlValidator;
     self.config.cssValidator = options.cssValidator;
-    self.config.noRecTrack = !!options.noRecTrack;
-    self.config.informativeOnly = !!options.informativeOnly;
+    self.config.noRecTrack = (options.noRecTrack === "true") ? true : false;
+    self.config.informativeOnly = (options.informativeOnly === "true") ? true : false;
     self.config.patentPolicy = options.patentPolicy;
     self.config.processDocument = options.processDocument;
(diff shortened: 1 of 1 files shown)


tripu:
BTW @deniak&hellip; well spotted!

PR #230, opened 2015-08-13
headers: allow trailing whitespaces in uris

### LINKED ISSUE #207: Incorrect "wrong syntax" error reported when link to this version has trailing spaces
If the spec to test contains trailing spaces in the `href` attribute of the "This version" and other "Previous version" links, Specberus reports errors such as:

* Wrong syntax for This Version link.
* error Short names differ between This and Latest Versions.
* error Short names differ between This and Previous Versions.

Example of markup that triggers the issue:

```html
<dl>
 <dt>This version:</dt>
 <dd><a href="http://www.w3.org/TR/2015/WD-presentation-api-20150610/ ">http://www.w3.org/TR/2015/WD-presentation-api-20150610/ </a>
 </dd>
 <dt>Latest published version:</dt>
 <dd><a href="http://www.w3.org/TR/presentation-api/ ">http://www.w3.org/TR/presentation-api/</a></dd>
 ...
</dl>
```

I don't have an example of an online spec because I'm going to update the version that I was testing. The thing is: Anolis generates this space, which is technically correct although a tad annoying perhaps, so it's not that easy to fix in the generated spec.

In the example, removing the final space before the closing double-quote in the `href` attribute fixes the issue. I suspect the code that would need to be tweaked is in `lib/rules/headers/dl.js`:

 https://github.com/w3c/specberus/blob/master/lib/rules/headers/dl.js#L87

In particular, adding a call to `trim()` should do the trick but note it would have to be done for the latest link and the previous link as well:

```javascript
var matches = ($linkThis.attr("href") || "").trim().match(new RegExp(vThisRex))
```

deniak:
Fix #207
Fix https://github.com/w3c/echidna/issues/201

### DIFF
--- a/lib/rules/headers/dl.js
+++ b/lib/rules/headers/dl.js
@@ -92,7 +92,7 @@ exports.check = function (sr, done) {
                        "\\/(\\d{4})\\/" +
                        (sr.config.status || "[A-Z]+") +
                        "-(.+)-(\\d{4})(\\d\\d)(\\d\\d)\\/?$";
-        var matches = ($linkThis.attr("href") || "").match(new RegExp(vThisRex))
+        var matches = ($linkThis.attr("href") || "").trim().match(new RegExp(vThisRex))
         ,   docDate = sr.getDocumentDate()
         ;
         if (matches) {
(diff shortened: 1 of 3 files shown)


tripu:
@deniak, are you sure this fixes w3c/echidna#201 too?

deniak:
@tripu you are right, echidna needs to be updated as well. I'll work on it

PR #236, opened 2015-08-18
Use 'published' rather than 'produced' to align with current respec-g…

dontcallmedom:
…enerated sotd

(I'll fix the rest of the text in ReSpec, but this one doesn't seem particularly warranted)


deniak:
That PR changes the text specified in [pubrules](http://www.w3.org/2005/07/pubrules?uimode=filter&uri=). If we are changing that text, all the specifications will be affected and we'll need comm/legal to step in.

dontcallmedom:
sorry, I only meant to change the Regex, not the patent policy text; will update

deniak:
@dontcallmedom LGTM but you forgot to update the test at the same time: https://github.com/w3c/specberus/blob/master/test/docs/headers/joint-publication.html#L68

PR #261, opened 2015-09-30
exception handling

plehegar:
Tried to follow the suggestions in
 https://github.com/w3c/specberus/pull/212

lib/rules/heuristic/shortname.js does however introduce duplicated code, so not sure if I like it...


lib/exceptions.js
```diff
@@ -0,0 +1,36 @@
+var records = require('./exceptions.json');
```
tripu: `'use strict';`
?

lib/validator.js
```diff
@@ -91,8 +94,13 @@ Specberus.prototype.validate = function (options) {
 };
 
 Specberus.prototype.error = function (rule, key, extra) {
-    this.sink.emit("err", rule, { key: key, extra: extra });
-};
+    if ((this.shortname !== undefined)
+        && (this.config.status === "WD" || this.config.status === "CR")
```
tripu: Why only for these two profiles?

lib/rules/heuristic/shortname.js
```diff
@@ -0,0 +1,49 @@
...
+    };
+
+    var $linkLate = extractLatestURI($dl);
+    if ($linkLate && $linkLate.length && $linkLate.attr("href") !== "") {
+      var lateRex = "^https?:\\/\\/www\\.w3\\.org\\/" + topLevel + "\\/(.+?)\\/?$"
+          ,   matches = ($linkLate.attr("href") || "").trim().match(new RegExp(lateRex));
+          if (matches) {
+             var sn = matches[1];
+             latestURI = $linkLate.text();
+             sr.metadata('shortname', sn);
```
tripu: If there's new metadata being emitted, please update also documentation in `README.md`.

lib/exceptions.js
```diff
@@ -0,0 +1,36 @@
...
+};
+
+Exceptions.prototype.has = function (shortname, rule, key, extra) {
+    var set = findSet(shortname);
+    if (set === undefined) return false;
+    for (var i = set.length - 1; i >= 0; i--) {
+        var exception = set[i];
+
+        return (rule === exception.rule
+                && compareValue(extra.type, exception.type));
```
tripu: `extra` may be undefined; this will throw an RTE.

lib/exceptions.js
```diff
@@ -0,0 +1,36 @@
...
+        if (count > 10)
+           return undefined;
+        var set = records[name];
+        count++;
+        if (typeof set === "String")
+            return _findSet(set);
+        return set;
+    }
+    return _findSet(shortname);
+}
```
tripu: Sorry, as `exceptions.json` is empty and there are no examples or test cases provided, I'm finding it difficult to understand the recursion here, and that limit depth of 10&hellip; What is that for?

This is what I entered in my `exceptions.json` file to try to mute the only error I get when [checking locally a WD that is already published](http://localhost:3000/?url=http%3A%2F%2Fwww.w3.org%2FTR%2Fencrypted-media%2F&profile=WD-Echidna&validation=simple-validation&noRecTrack=false&informativeOnly=false&echidnaReady=true&patentPolicy=pp2004&processDocument=2015):

```json
{
  "encrypted-media": [
    {"rule": "echidna.todays-date"}
  ]
 }
```

Is that valid syntax?

lib/exceptions.js
plehegar: Added by https://github.com/w3c/specberus/commit/86cc5b92c8f06130fedf8203c874888c50584b38

lib/validator.js
plehegar: Because we don't want to allow exceptions beyond CR, ie for PR and REC.

lib/rules/heuristic/shortname.js
plehegar: See https://github.com/w3c/specberus/commit/8cc80858ff63fde4ec126ea664a2e5451b4b19ea

lib/exceptions.js
plehegar: See https://github.com/w3c/specberus/pull/212
Here is an example of valid syntax:
```javascript
 "hr-time-2": [
    {
      "key": "error",
      "name": "validation.css"
    },
    {
      "key": "error",
      "extra":
      {
        "message": "Attribute for not allowed on element a"
      },
      "name": "validation.html"
    }
  ]
}
```

tripu:
@plehegar:
I reviewed your latest changes, and tested them with one spec. Those issues/doubts that I saw seem OK now, thank you!

But I also re-read the last comments in [the thread of your original PR](https://github.com/w3c/specberus/pull/212). The problems raised there led us to consider instead using the structured output of the CSS validator (in JSON format), so that we would avoid the risks and the fragility of matching UI strings etc. See [this comment](https://github.com/w3c/specberus/pull/212#issuecomment-134929186) and [this other comment](https://github.com/w3c/specberus/pull/212#issuecomment-134935698).

But that is exactly what this is PR is doing again, isn't it? You're not using the structured output of the CSS validator in particular, but cancelling only the errors that Specberus knows about, and still using localised messages for that&hellip;

Am I missing anything?

(BTW, I've started [a wiki page about this exception mechanism](https://github.com/w3c/echidna/wiki/Exceptions); even if it's not deployed yet. Feel free to edit.)

PR #264, opened 2015-11-03
New profile: First Public Candidate Recommendation

deniak:
@plehegar that MR adds a new profile (`First Public Candidate Recommendation`) to specberus according to process 2015.
Let me know if you are ok the name of the profile (appearing in the `<h2>`).


tripu:
@deniak,
I committed a tiny little update to the README here.
I guess this profile does not exist as such in the old pubrules checker, so no need to extract the messages and add them to `lib/rules.json` like the rest&hellip;?

@plehegar, this is ready for your review :¬)

PR #272, opened 2015-12-01
Ignore at-rule errors from css-ui-3

plehegar:
Based on css errors from the latest CSS UI CR.


deniak:
see http://jigsaw.w3.org/css-validator/validator?profile=css3&uri=http://www.w3.org/TR/2015/CR-css-variables-1-20151203/

tripu:
It does exactly what it says on the tin.

Issue #187, opened 2015-04-16
Support for copyright exceptions


dontcallmedom:
There a a (arguably somewhat limited) number of specs that use a different copyright statement from the one blessed by pubrules, and these exceptions have been approved by the power that be.

At the moment, specberus reject these exceptions; to make these document echidna-publishable, it'd be great if they could be blessed by specberus too.

I have two specific use cases for my group:

* http://w3c.github.io/webrtc-pc/
* http://w3c.github.io/mediacapture-main/getusermedia.html

darobin:
I agree that that would be useful, but how do you propose to implement it? Just writing down some thoughts here.

We could make a specific option for Specberus to ignore copyright errors. That's the simplest to implement, but there is a risk that we could start adding a lot of exceptions which would be unpleasant to manage.

We could take it one notch up and allow Specberus to receive a list of errors to ignore (using the key names for errors). The code would be relatively simple, the UI a bit more painful.

For both of the above Echidna would need a way of being informed that an exception has been granted for a document, and transmit that information to Specberus. I'm guessing that would be registered with the token.

Finally, there's the option of addressing this in Echidna alone. Echidna knows (or can know) which specific errors it is getting from Specberus. It could, on its own, decide to ignore some of them.

tripu:
I fear that we'll implement different kinds of flexible mechanisms to introduce exceptions in many aspects of specs (not only the copyright statement), and that that will be prone to errors and make the system more difficult to maintain and update.

@dontcallmedom, these are exceptions in the sense that only those two specs are allowed to be published with that copyright statement, right?

@darobin's idea about muting specific error messages for specific tokens sounds good in principle. My main concern is that ignoring copyright error messages will not just allow those editors to use that special statement&nbsp;&mdash;&nbsp;it will effectively let them to use *any* copyright statement&nbsp;&mdash;&nbsp;or none at all! This may be an issue whenever we ignore on purpose any error message from Specberus, incidentally. So the &ldquo;exceptions system&rdquo; should be smarter than that: instead of saying &ldquo;for spec S, just ignore any error messages about X&rdquo; we should be able to say &ldquo;for spec S, instead of checking X1, check X2&rdquo;. In other words, overwriting entire rules in Specberus, per individual spec/token&hellip;

Is it as complex as I'm painting it here?

plehegar:
@dontcallmedom note that we no longer use ther terms "All Rights Reserved." in copyright statements btw

deniak:
See #208

[...]

PR #273, opened 2015-12-02
Document approved exceptions to copyright statement for two WebRTC specs

dontcallmedom:
I assume this is the right syntax for copyright extensions based on what I've gathered from https://github.com/w3c/specberus/issues/187

### DIFF
--- a/lib/exceptions.json
+++ b/lib/exceptions.json
@@ -1,12 +1,12 @@
 {
  "webrtc": [
     {
-      "rule": "headers.copyright.no-match"
+      "rule": "headers.copyright"
     }
   ],
  "mediacapture-streams": [
     {
-      "rule": "headers.copyright.no-match"
+      "rule": "headers.copyright"
     }
   ]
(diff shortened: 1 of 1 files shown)


lib/exceptions.json
```diff
@@ -1,2 +1,12 @@
 {
+ "webrtc": [
+    {
+      "rule": "headers.copyright.no-match"
```
deniak: The name of the rule is `headers.copyright`. The current exception system is not granular enough to skip specific errors raised by a rule.

dontcallmedom: Too bad, I was hoping specberus could still warn us on other copyright related stuff (e.g. year in copyright, correct links, etc).
Time permitting, I might look at making the exception system more granular, but in the meantime, I've updated the PR

deniak: I created an issue https://github.com/w3c/specberus/issues/274 (cc @plehegar)

PR #275, opened 2015-12-02
Added more flexibility in exceptions

plehegar:
This allows the following definitions:

```json
{
  "css-": [
    {
      "rule": "validation.css",
      "type": "at-rule"
    }
  ],
  "css-ui-3": [
    {
      "rule": "validation.css",
      "type": "generator.unrecognize"
    }
  ]
}
```

Based on startsWith comparison, any shortname starting with "css-" will match the set of rules, and
css-ui-3 will take into account both set of rules (so both errors will be excluded in that case).

### DIFF
--- a/lib/exceptions.js
+++ b/lib/exceptions.js
@@ -6,26 +6,24 @@ function compareValue(data, ref) {
     return (ref === undefined) || (data === ref);
 }
 
-function findSet(shortname) {
-    var count = 0;
-    function _findSet(name) {
-        if (count > 10)
-           return undefined;
-        var set = records[name];
-        count++;
-        if (typeof set === "String")
-            return _findSet(set);
-        return set;
+// from all of the arrays in the set of rules, find the ones applicable
+// according to a shortname
+function getApplicableSet(shortname) {
+    var set = [];
+    for (var key in records) {
+      if (shortname.startsWith(key)) {
+        set = set.concat(records[key]);
+      }
     }
-    return _findSet(shortname);
+    return set;
 }
 
 var Exceptions = function () {
 };
 
+// returns true if one of the rules matches the shortname, rule, and extra.type
 Exceptions.prototype.has = function (shortname, rule, key, extra) {
-    var set = findSet(shortname);
-    if (set === undefined) return false;
+    var set = getApplicableSet(shortname);
     for (var i = set.length - 1; i >= 0; i--) {
         var exception = set[i];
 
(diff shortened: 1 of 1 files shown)


deniak:
I don't think it's safe to have a set of exceptions for all the specifications starting with a specific string.
We are taking the risk to skip rules for a specification without even noticing it. There will be new shortnames in the future and who knows, another WG might want a shortname starting with `css-` too.
As its name suggests, exceptions should be *exceptional*.

I understand it makes sense for the CSS WG to have such system in place as all their specifications are affected by the CSS validator not being up-to-date but I think the validator should trigger a warning instead of an error if that's the case (eg. [css-ui-3](http://jigsaw.w3.org/css-validator/validator?uri=http%3A%2F%2Fwww.w3.org%2FTR%2F2015%2FCR-css-ui-3-20150707%2F&profile=css3&usermedium=all&warning=1&vextwarning=&lang=en&output=json)).

plehegar:
the at-rule exception is recurring on ALL CSS drafts. How do you propose to codify that without repeating it 50 times?

deniak:
My suggestion is to update the CSS validator so it simply reports a warning instead of an error (cc @ylafon)

plehegar:
The error is triggered by the use of @supports , which is defined in
 http://www.w3.org/TR/css3-conditional/#at-supports

plehegar:
See also http://caniuse.com/#feat=css-featurequeries

plehegar:
An other information bit: the official definition of CSS is in
 http://www.w3.org/TR/CSS/#css
and it includes @supports

tripu:
(Description edited, but only to format that JSON.)

tripu:
I think I agree with @deniak on this one: we're opening the door to these exception not being exceptional at all if we allow shortname prefixes (and it's risky, as he pointed out). We do that, and soon we find ourselves allowing wildcards for validation errors, too. It'll make it almost too easy to circumvent and ignore problems in specs.

I'd rather have a list of exceptions with 50 lines like these in it

```json
{
  "css-foo": [{"rule": "validation.css", "type": "at-rule"}],
  "css-bar": [{"rule": "validation.css", "type": "at-rule"}],
  "css-baz": [{"rule": "validation.css", "type": "at-rule"}]
}
```

to make it explicit that we have granted fifty exceptions to fifty specs.

Anyway, hopefully these changes in the CSS validator will be feasible soon; let's hear what @ylafon has to say!

deniak:
I talked to @ylafon and switching from error to warning is not an easy task. He said he'd rather spend time implementing css3-conditional

deniak:
@plehegar, I just merged #294 so I think we can close that one. Sorry I should have continued with your PR.

Issue #267, opened 2015-11-12
2016 TR style


deniak:
Upcoming changes so far:
* MUST use `/TR/StyleSheets/2016/base.css`
* MUST use `nav.toc`
* MUST use meta view port
* MAY use `<p role=navigation id="back-to-top">`

 see http://fantasai.inkedblade.net/style/design/w3c-restyle/2016/sample

--
\[List below added by @tripu, 18 Jan 2016\]

- MUST
  - [x] #280 (`<meta name="viewport">`)
  - [x] #281 (`<nav id="toc">`)
  - [x] #282 (CSS from `w3.org/StyleSheets/TR/2016/`)
  - [x] #283 (link to `fixup.js`)
  - [x] #292 (new W3C logo `//www.w3.org/StyleSheets/TR/2016/logos/W3C`)
- SHOULD
  - [x] #284 (use `<ol class="toc">` for the table of contents)
  - [x] #285 (wrap section numbers (in the TOC and in headings) in `<span class="secno">`)
  - Emit a warning if the spec is *not* written in HTML5 (#281 kind of does that; although ideally it should be a top-level warning, well visible and pointing to some mail message or page explaining that support for doctypes other than HTML5 will be abandoned during 2016)
- MAY
  - [x] #288 Append back-to-top link at end of document:  
`<p role=navigation id="back-to-top"><a href="#toc"><abbr title="Back to Top">&uarr;</abbr>`
  - [x] #289, which includes:
    - [Use `.example`, `.def`, `.note`, `.marker`, and other special box markup](http://fantasai.inkedblade.net/style/design/w3c-restyle/2016/sample#boxes)
    - [Use `.index` for index lists and tables](http://fantasai.inkedblade.net/style/design/w3c-restyle/2016/sample#indexing)
    - Make sure spec illustrations/tables fit within an A4 sheet of paper when printed; use `.overlarge` class if needed to improve experience on-screen for things that overflow (but note that they might still be issues)

tripu:
Assigning to @plehegar because the next step is on him to give a green light to TR styles 2016.

tripu:
Done.

PR #280, opened 2016-01-17
Check new tag <meta name="viewport">

tripu:
cf. #267.

### DIFF
--- a/lib/l10n-en_GB.js
+++ b/lib/l10n-en_GB.js
@@ -166,6 +166,8 @@ for extended information about the publication rules.'
     // style/sheet
 ,   "style.sheet.last":         "W3C TR style sheet must be last."
 ,   "style.sheet.not-found":    "Missing W3C TR style sheet."
+    // style/meta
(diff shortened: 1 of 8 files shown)


deniak:
LGTM but I wonder if instead of looking for a specific value of the `content` attribute, we shouldn't [parse it](https://drafts.csswg.org/css-device-adapt/#parsing-algorithm) and check each property.
Currently, the following attributes are failing:
- `content='width=device-width,initial-scale=1,shrink-to-fit=no'`
- `content='initial-scale=1, width=device-width, shrink-to-fit=no'`

tripu:
OK, @deniak. I think this check is no worse than those other checks we have (case-sensitive, verbatim text, even blanks are fixed).

But I'll do some simple parsing of the properties, as you said!

dontcallmedom:
in case this helps, I have a [library that parses meta viewport content attributes](https://github.com/dontcallmedom/metaviewport-parser).

tripu:
@deniak: I used @dontcallmedom's package (cheers!) for proper parsing; I think the check is better now.

OK to merge before 1 March?

deniak:
@tripu it's ready to be merged. Nice work.

PR #281, opened 2016-01-17
Check new navigation element

tripu:
`<nav id="toc">` (HTML5) preferred.
Only `<div id="toc">` ~~or `<section id="toc">`~~ allowed as an alternative for now, but ~~they throw~~ it throws a warning.

cf. #267.


lib/rules/headers/h2-toc.js
```diff
@@ -5,18 +5,35 @@ exports.name  = 'headers.h2-toc';
...
+    var EXPECTED_HEADING = /table\s+of\s+contents/i;
+    var toc = sr.$('nav#toc > h2');
+
+    if (1 === toc.length) {
+        if (!EXPECTED_HEADING.test(sr.$(toc[0]).text())) {
+            sr.error(exports.name, 'not-found');
+        }
+    }
+    else {
+        toc = sr.$('#toc > h2');
```
deniak: From the discussion with @plehegar and Fantasai, we said we would either accept `<nav>` or `<div>`. I don't think we should allow other tags.

lib/l10n-es_ES.js
```diff
@@ -21,7 +21,9 @@ exports.messages = {
 ,   "headers.h2-status.no-h2":  "No status+date h2 element."
 ,   "headers.h2-status.bad-h2": "Incorrect status h2 header."
     // headers/h2-toc
-,   'headers.h2-toc.missing':  'There is no table of contents after the status section, labeled with a <code>&lt;h2&gt;</code> element with content <em>&ldquo;Table of Contents&rdquo;</em>.'
+,   'headers.h2-toc.not-found':  'There is no table of contents inside a navigation element (<code>&lt;nav id="toc"&gt; &hellip; &lt;h2&gt;Table of Contents&lt;/h2&gt; &hellip; &lt;/nav&gt;</code>).'
+,   'headers.h2-toc.too-many':   'Found more than one navigation elements; possibly containing the table of contents. There should be just one <code>&lt;nav id="toc"&gt;</code>.'
+,   'headers.h2-toc.not-html5':  'Support for old HTML doctypes is deprecated and will be abandoned soon; please upgrade to HTML5. The expected navigation element is: <code>&lt;nav id="toc"&gt;</code>.'
```
deniak: That message refers to the rule `headers.h2-toc`. That's what the user currently sees on the interface. I'm afraid he will get confused by these messages referring to the navigation.

deniak:
@tripu can you update the existing test drafts instead of creating new ones? We are starting to have a lot of test drafts used only by one test.

lib/rules/headers/h2-toc.js
tripu: I thought so, too, @deniak.

To confirm, I checked the latest published specs ([example](https://www.w3.org/TR/2016/WD-webdriver-20160120/), [example](https://www.w3.org/TR/2016/WD-preload-20160119/), [example](https://www.w3.org/TR/2016/WD-vehicle-information-api-20160119/)) and saw that all of them are using `<section id="toc">`. Then I saw that [@fantasai is not enforcing one particular element in her CSS; only matching `#toc` (and she actually wrote `<section id="toc">` in the comments)](https://github.com/w3c/tr-design/blob/gh-pages/src/base.css#L9), so I thought we were fine with `section` too&hellip; That's why I'm matching only `#toc` here.

@fantasai, @plehegar: what needs an update? This Specberus rule, or the CSS file?

tripu:
You're right about the sample documents, @deniak. I'll merge this new one into one of the existing ones.

lib/rules/headers/h2-toc.js
fantasai: The CSS file needs an update. <nav> is clearly the better element to use. :)

tripu:
@deniak, @fantasai: I updated the check to accept only `<nav>` (preferred) or `<div>` (throws a warning). I wonder how many editors will continue submitting specs with `<section>`, though&hellip; :confused: (this change is explained clearly on @plehegar's message to spec-prod, though).

@deniak: I removed the two new samples, and modified existing ones instead to test the rule.

> That message refers to the rule `headers.h2-toc`. That's what the user currently sees on the interface. I'm afraid he will get confused by these messages referring to the navigation.

I don't understand this comment. TOC and navigation are the same thing&nbsp;&mdash;&nbsp;we're only changing the pattern we look for on the document. Right?

deniak:
:+1:

PR #283, opened 2016-01-17
Check new expected script, “fixup.js”

tripu:
Notes/questions (don't merge until these are addressed):

- [x] **About the rule**: this rule *checks that the script is linked on the spec already*, ie it does *not* add a link to the script. This follows what we usually do with documents (we don't alter them, but point editors to errors and ask them to fix). Let's discuss if this is is going to pose a problem and we really need to append the `<script>` element ourselves.  
&rarr; **Update (19 Jan)**: confirmed after chatting with @deniak and @plehegar that we're *checking* this, not *adding* it to the spec.
- [x] **About the URL of the script**: @plehegar mentioned `https://www.w3.org/scripts/fixup.js` in the last e-mail thread. But I'd prefer not to have scripts directly under that directory (`scripts`). Since we'll deploy the new CSS at `w3.org/StyleSheets/TR/2016/`, this PR suggests we put this script at the analogous location, `w3.org/scripts/TR/2016/fixup.js` (so, `w3.org/scripts/TR/` will be used only for TR scripts going forward). Sounds good?  
&rarr; **Update (16 Feb)**: confirmed the location; cf. [`tabatkins/bikeshed#569#issuecomment-184607100`](https://github.com/tabatkins/bikeshed/issues/569#issuecomment-184607100).

@TODO: **[the script](https://github.com/w3c/tr-design/blob/gh-pages/src/fixup.js) has to be deployed to that location before this milestone** (29 Jan 2016).

cf. #267.

### DIFF
--- a/lib/l10n-en_GB.js
+++ b/lib/l10n-en_GB.js
@@ -170,6 +170,8 @@ for extended information about the publication rules.'
 ,   "style.sheet.not-found":    "Missing W3C TR style sheet."
     // style/meta
 ,   "meta.not-found":           "The viewport meta tag is required (<code>&lt;meta name=\"viewport\" content=\"width=device-width, initial-scale=1, shrink-to-fit=no\"&gt;</code>)."
+    // style/script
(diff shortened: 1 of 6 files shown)


tripu:
Ready to be merged before 1 Mar 2016.

deniak:
:+1:

PR #284, opened 2016-01-18
Check new expected element <ol class="toc">

tripu:
NB: throws only a warning if not found, ie this rule does *not* fail.

cf. #267.

### DIFF
--- a/lib/l10n-en_GB.js
+++ b/lib/l10n-en_GB.js
@@ -24,6 +24,9 @@ exports.messages = {
 ,   'headers.h2-toc.not-found':  'There is no table of contents inside a navigation element (<code>&lt;nav id="toc"&gt; &hellip; &lt;h2&gt;Table of Contents&lt;/h2&gt; &hellip; &lt;/nav&gt;</code>).'
 ,   'headers.h2-toc.too-many':   'Found more than one navigation elements; possibly containing the table of contents. There should be just one <code>&lt;nav id="toc"&gt;</code>.'
 ,   'headers.h2-toc.not-html5':  'Support for old HTML doctypes is deprecated and will be abandoned soon; please upgrade to HTML5. The expected navigation element is: <code>&lt;nav id="toc"&gt;</code>.'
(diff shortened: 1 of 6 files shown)


lib/l10n-en_GB.js
```diff
@@ -22,6 +22,8 @@ exports.messages = {
 ,   "headers.h2-status.bad-h2": "Incorrect status h2 header."
     // headers/h2-toc
 ,   'headers.h2-toc.missing':  'There is no table of contents after the status section, labeled with a <code>&lt;h2&gt;</code> element with content <em>&ldquo;Table of Contents&rdquo;</em>.'
+    // headers/ol-toc
+,   'headers.ol-toc.not-found':  'The TOC should be an <code>&lt;ol class="toc"&gt;</code> inside the main navigation element (<code>&lt;nav id="toc"&gt;</code>).'
```
deniak: For the keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", I suggest we use `<span class="rfc2119">keyword</span>` like it appears in the old pubrules.

lib/rules/headers/ol-toc.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Check the presence of the actual TOC (<code>&lt;ol class="toc"&gt;</code>) inside the main navigation element.
+ */
+
+exports.name  = 'headers.ol-toc';
+
+exports.check = function (sr, done) {
+
+    var toc = sr.$('#toc ol.toc');
```
deniak: Search for `nav#toc` or `div#toc`

tripu: Well spotted!
Done.

lib/l10n-en_GB.js
tripu: This I'll address separately in #293

tripu:
Ready to merge, @deniak.

deniak:
:+1:

PR #285, opened 2016-01-18
Check presence of <span class="secno"> in TOC/headings

tripu:
Kind of *heuristically*.
NB: throws only a warning if no instances found, ie this rule does *not* fail.

cf. #267.

### DIFF
--- a/lib/rules/headers/secno.js
+++ b/lib/rules/headers/secno.js
@@ -8,7 +8,7 @@ exports.name  = 'headers.secno';
 
 exports.check = function (sr, done) {
 
-    var secnos = sr.$('h1 span.secno, h2 span.secno, h3 span.secno, h4 span.secno, h5 span.secno, h6 span.secno, #toc span.secno, .toc span.secno');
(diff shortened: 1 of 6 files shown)


lib/rules/headers/secno.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Check (more or less) if items in the TOC and headings are wrapped inside <code>&lt;span class="secno"&gt;</code>.
+ */
+
+exports.name  = 'headers.secno';
+
+exports.check = function (sr, done) {
+
+    var secnos = sr.$('h1 span.secno, h2 span.secno, h3 span.secno, h4 span.secno, h5 span.secno, h6 span.secno, #toc span.secno, .toc span.secno');
```
deniak: With #281 ready to be merged, I don't think there should be `.toc span.secno` in the list. `#toc span.secno` is enough.

tripu: Done, @deniak.

deniak:
:+1:

PR #288, opened 2016-01-19
Check expected back-to-top link

tripu:
NB: throws only a warning if not found, ie this rule does *not* fail.

cf. #267.


deniak:
In the test file `test/docs/headers/back-to-top.html`, remove unnecessary markup and keep it minimal so it's easy to understand what you are trying to test like in `test/docs/headers/logo.html`

tripu:
Done, @deniak.

deniak:
:+1:

PR #289, opened 2016-01-19
Emit new informative messages regarding recommended markup

tripu:
NB: adds 3 new `info` messages to the output; this rule does *not* fail.

cf. #267.


deniak:
:+1:

PR #294, opened 2016-02-26
Generic exceptions

deniak:
`exceptions.json` now takes regexp for shortnames.

### DIFF
--- a/lib/exceptions.json
+++ b/lib/exceptions.json
@@ -1,15 +1,27 @@
 {
-  "webrtc": [
+  ".*": [
+    {
+      "rule": "validation.css",
+      "type": "generator.unrecognize",
+      "source": "https://www.w3.org/2016/01/base.css"
+    },
+    {
+      "rule": "validation.css",
+      "type": "generator.unrecognize",
+      "source": "https://www.w3.org/StyleSheets/TR/2016/base.css"
+    }
+  ],
+  "^webrtc$": [
     {
       "rule": "headers.copyright"
     }
   ],
-  "mediacapture-streams": [
+  "^mediacapture-streams$": [
     {
       "rule": "headers.copyright"
     }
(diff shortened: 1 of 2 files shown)


plehegar:
tested and looks good to me.

lib/exceptions.json
```diff
@@ -1,15 +1,22 @@
...
+      "source": "https://www.w3.org/StyleSheets/TR/2016/base.css"
+    }
+  ],
+  "^webrtc$": [
     {
       "rule": "headers.copyright"
     }
   ],
-  "mediacapture-streams": [
+  "^mediacapture-streams$": [
```
tripu: I thought that `-` here would be [mis]interpreted as defining a range of characters.
But I just learnt that [hyphens do not need to be escaped when they don't appear inside square brackets](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#character-sets).
Still, maybe `"^mediacapture\-streams$"` to err on the safe side, and to remind anyone who is editing the exceptions file that these metacharacters *usually* ought to be escaped&hellip;?

lib/exceptions.json
```diff
@@ -1,15 +1,22 @@
...
     {
       "rule": "headers.copyright"
     }
   ],
-  "css-ui-3": [
+  "^css-ui-3$": [
```
tripu: Ditto.

tripu:
@deniak: tiny suggestion inline. Otherwise, :+1: to rebase &amp; merge.

deniak:
@tripu, since the exceptions are listed in a json file, we would need to use `\\` to escape the hyphens. I think it would add unnecessary noise to the file.

tripu:
Ah, good point, @deniak. OK then!

PR #297, opened 2016-03-03
Fix #296 (better detection of "fixup.js")

### LINKED ISSUE #296: https? protocol for fixup.js
specberus rejects links to  https://www.w3.org/scripts/TR/2016/fixup.js

instead, it expects :

  var candidates = sr.$("script[src='//www.w3.org/scripts/TR/2016/fixup.js']");

shouldn't we say (https?:)? at least? (like it's done for the logo)

tripu:


lib/rules/style/script.js
```diff
@@ -8,9 +8,18 @@ exports.name = 'style.script';
 
 exports.check = function (sr, done) {
 
-    var candidates = sr.$("script[src='//www.w3.org/scripts/TR/2016/fixup.js']");
+    const PATTERN_SCRIPT = /^(https?:)?\/\/(www\.)?w3\.org\/scripts\/tr\/2016\/fixup\.js$/i;
```
deniak: it should be `(https:)?`

tripu:
@Mouvedia, all documents [should be HTML5](https://www.w3.org/pubrules/doc/rules/?profile=FPWD#valideHTML) (/cc @deniak).

Mouvedia:
@tripu https://www.w3.org/TR/html5/introduction.html#html-vs-xhtml

XHTML documents ought to be valid and not all specifications are non-XML documents.

tripu:
@Mouvedia, [all specs recently published](https://www.w3.org/TR/) are `text/html`, not `application/xhtml+xml`.
I still think it makes sense to [document this requirement with the most common chunk of code](https://www.w3.org/pubrules/doc/rules/?profile=FPWD#fixupJs), and there is no need to cover all possible scenarios.

You said all specs you checked are affected by this.
What are those specs, and how exactly are they affected?
Thank you.

Mouvedia:
> What are those specs, and how exactly are they affected?

https://www.w3.org/standards/xml/core#specifications

tripu:
@Mouvedia: sorry, I still don't understand what specification in particular is &ldquo;affected&rdquo; by this PR, and how it is a problem. Can you point to a spec or two, and elaborate?

Mouvedia:
> Can you point to a spec or two, and elaborate?

Here are 2 examples taken from the link that Iv given you:

- https://www.w3.org/TR/2009/REC-xml-names-20091208/
- https://www.w3.org/TR/2008/REC-xml-20081126/

Both are XHTML 1.0 transitional and had `<script src="https://www.w3.org/scripts/TR/fixup.js"></script>` added without taking into account that it would render them invalid (if they weren't already—some were).

deniak:
@Mouvedia This is indeed a bug introduced when we added that script. All the XHTML documents should now be fixed. Thanks for your feedback.

PR #299, opened 2016-03-17
Add method for metadata extraction

### LINKED ISSUE #34: Auto detect profile from URL

tripu:
This PR adds a method `extractMetadata()` to `Specberus` that returns:

* Detected profile (fixes #34).
* Detected deliverers:
  * Names.
  * URL of their homepage.

Also:

* It adds code and doc samples to test metadata extraction against ~~7~~ 6 actual specs (`test/samples.json`, `test/docs/metadata/*`).
* Tests (old and new) are split into different files under `tests/` now.
* Refactor a few routines into separate modules (sink, utils).

**TO-DO:**

* ~~Return ID of deliverers too (necessary for w3c/echidna#255).~~ **Done.**
* ~~Remove traces of work on a REST API (that'll go on a separate PR, and it's very low-priority right now).~~ **Done (see #300).**
* After `extractMetadata()`, save results in a dictionary, associated with the source. That way, when `validate()` runs immediately after that, we save that step. It will be also necessary both on the web UI and for Echidna. **Not really necessary for what we need to accomplish here; Echidna will call `extractMetadata()` first, and then use the value of `detectedProfile` to call `validate()` with the right options.**
* ~~Update documentation to reflect changes in public methods and their signatures, and the move from the old metadata pieces being emitted to the new `extractMetadata()` method.~~ **Done.**

**Questions:**

* Do we want to test metadata extraction against public URLs, or locally against saved copies of those specs? Right now we have both; it's decided depending on the environment (Travis/local). **Decided w/ @deniak to keep the current &ldquo;hybrid&rdquo; approach (public URLs in general + local docs for faster tests in Travis CI), but trimming the local samples to remove their content, which is not relevant. Done.**


app.js
```diff
@@ -1,24 +1,38 @@
...
+
+// Internal packages:
+const package = require('./package.json')
+,   api = require('./lib/api')
+,   l10n = require('./lib/l10n')
+,   sink = require('./lib/sink')
+,   validator = require('./lib/validator')
+;
+
+const app = express()
```
deniak: These are not constants.

tripu: Yes, they are. We don't assign a different value to these variables after these declarations.

deniak: what about `profiles`?

README.md
```diff
@@ -75,6 +76,34 @@ following fields:
 * `events`: An event sink which supports the same interface as Node.js's `EventEmitter`. Required. See
   below for the events that get generated.
 
+### `extractMetadata(options)`
+
+This method returns a simple object with metadata inferred from the document.
+The `options` accepted are equal to those in `validate()`, except that a `profile` is not necessary and will be ignored (finding out the profile is one of the
+goals of this method).
+
+The returned `Object` may contain up to 3 properties: `detectedProfile`, `detectedDeliverers` and `detectedDelivererIDs`.
```
deniak: `detectedDeliverers` and `detectedDeliverersIDs` are redundant and may lead to inconsistencies. I'd say keep `detectedDelivererIDs` only. From the id, the name and the homepage of the group are easy to get using the API

app.js
deniak: It's not entirely true, see `profiles` for instance but I see `const` doesn't protect the object attributes.

lib/validator.js
```diff
@@ -29,6 +31,51 @@ Specberus.prototype.clearCache = function () {
...
+        if (err) return self.throw(err);
+        self.$ = query;
+        self.sink.emit("start-all", profileMetadata);
+        var total = (profileMetadata.rules || []).length
+        ,   done = 0
+        ;
+        profileMetadata.rules.forEach(function (rule) {
+            rule.check(self, function (result) {
+                if (result) {
+                    // console.dir(result);
```
deniak: remove comment

lib/rules/heuristic/group.js
```diff
@@ -1,45 +0,0 @@
-
```
deniak: That rule cannot be removed. We need to list the detected deliverers on the interface so the user can check if everything is ok

lib/rules/metadata/profile.js
```diff
@@ -0,0 +1,48 @@
...
+
+    sr.$(SELECTOR_SUBTITLE).each(function() {
+        candidate = sr.norm(sr.$(this).text()).toLowerCase();
+        i = 0;
+        while (i < profiles.tracks.length) {
+            track = profiles.tracks[i].profiles;
+            j = 0;
+            while (j < track.length) {
+                profile = track[j];
+                if (-1 !== candidate.indexOf(profile.name.toLowerCase()) && matchedLength < profile.name.length) {
```
deniak: `profile.name` will return `WD`, `FPWD`, etc.
I think you want to use `profile.config.longStatus`

tripu: No, this is using `profiles.json`; the long name is called `name` there.

app.js
tripu: Yeah, we just make sure they're constants here, not immutable.

lib/validator.js
tripu: Oops!

README.md
```diff
@@ -75,6 +76,24 @@ following fields:
...
+goals of this method).
+
+The returned `Object` may contain up to 2 properties: `detectedProfile` and `detectedDelivererIDs`.
+If some of these pieces of metadata cannot be deduced, that key will not exist, or its value will not be defined.
+
+An example:
+
+```json
+{
+  "detectedProfile": "WD",
```
deniak: better remove 'detected' in the properties name.

lib/rules/heuristic/group.js
tripu: Chatted with @deniak: the new method `extractMetadata()` will be used instead of these old rules that return just `info()` messages.

README.md
tripu: Done.

tripu:
@deniak, I addressed your last comment. The ball's in your court again.

PR #300, opened 2016-03-20
Set up REST API

tripu:
This PR sets up a REST API using Express, with 3 methods:

* `version`
* `metadata`
* `validate`

And adds a test file for Mocha with placeholders for API tests.

Because this functionality was moved here from branch `tripu/metadata`, there are a few files in common (`sink.js`, `util.js`, `samples.json`). Those should be OK when we merge the branches.

As the API relies on the new method `Specberus.extractMetadata()`, **do *not* merge this before #299**; it won't work.

**TO-DO:** [Updated 3/May/2016]

* ~~Write the tests. My plan is to launch the Specberus server locally first, then use `superagent` to throw a few requests at it, and check the results.~~ :ballot_box_with_check:
* ~~POST or GET? Method `version` should be GET. I'm not sure about the other two.~~ :ballot_box_with_check: All are GET.
* ~~Refactor if/else block for `api/metadata` and `api/validate`: the signature and checks are almost identical for both.~~ :ballot_box_with_check:

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ The interface you get when you `require("specberus")` is that from `lib/validato
 `Specberus` instance that is properly configured for operation in the Node.js environment
 (there is nominal support for running Specberus under other environments, but it isn't usable at this time).
 
-(See also [the REST API](#rest-api).)
+(See also [the REST API](#5-rest-api).)
 
 ### `validate(options)`
 
@@ -132,7 +132,7 @@ This is an example of the value of `Specberus.meta` after the execution of `Spec
 
 ## 5. REST API
 
-Similar to the [JS API](#js-api), Specberus exposes a REST API via HTTP too.
+Similar to the [JS API](#4-js-api), Specberus exposes a REST API via HTTP too.
 
 The endpoint is `<host>/api/`.
 Use either `url` or `file` to pass along the document (neither `source` nor `document` are allowed).
--- a/README.md
+++ b/README.md
@@ -142,6 +142,8 @@ Extract all known metadata from document; [returns a JSON object with inferred p
 
 Check the document ([syntax](#validateoptions)); fails and returns an array of errors, or succeeds and returns the profile.
 
+The special profile `auto` is also available.
+
 ### Examples
 
 * `<host>/api/version`
--- a/README.md
+++ b/README.md
@@ -123,10 +123,10 @@ This is an example of the value of `Specberus.meta` after the execution of `Spec
 
 ## REST API
 
-Similar to the [JS API](#js-api), Specberus exposes a REST API too.
+Similar to the [JS API](#js-api), Specberus exposes a REST API via HTTP too.
 
 The endpoint is `<host>/api/`.
-Use either `url` or `file` to pass along the document (`source` and `document` are not allowed).
+Use either `url` or `file` to pass along the document (neither `source` nor `document` are allowed).
 
 There are three `GET` methods available.
 
@@ -136,11 +136,12 @@ Returns the version string, eg `1.5.3`.
 
 ### `metadata`
 
-Extract all known metadata from document; [returns a JSON object with inferred properties](#extractmetadataoptions).
+Extract all known metadata from a document; see [below](#return-values) for information about the return value.
 
 ### `validate`
 
-Check the document ([syntax](#validateoptions)); fails and returns an array of errors, or succeeds and returns the profile.
+Check the document ([syntax](#validateoptions)).
+Many of [the options understood by the JS method `validate`](#validateoptions) are accepted.
 
 The special profile `auto` is also available.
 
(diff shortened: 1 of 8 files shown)


tripu:
Fixes #320.

tripu:
@deniak, ready for review.
(The 2 errors in the test suite aren't specific about the API; they are caused by #347.)

lib/api.js
```diff
@@ -0,0 +1,91 @@
...
+ * Build an "options" object based on an HTTP query string.
+ *
+ * @param {Object} query - an HTTP request query.
+ * @returns {Object} an "options" object that can be used by Specberus.
+ */
+
+const parseSource = function(query) {
+    const result = {};
+    if (query.url) result.url = query.url;
+    else if (query.source) result.source =uery.source;
```
deniak: s/uery/query/

README.md
```diff
@@ -119,6 +121,25 @@ This is an example of the value of `Specberus.meta` after the execution of `Spec
...
+
+* `version`: returns the version string, eg `1.5.3`.
+* `metadata`: extract all known metadata from document; [returns JSON object with inferred properties](#extractmetadataoptions).
+* `validate`: check the document ([syntax](#validateoptions)); fails and returns an array of errors, or succeeds and returns the profile.
+
+Examples:
+
+* `<host>/api/version`
+* `<host>/api/metadata?url=http://example.com/doc.html`
+* `<host>/api/validate?file=/home/me/docs/spec.html`
```
deniak: That method ('file' QS param) won't work unless the file is already on the server which won't happen. Same thing for 'document'.
Regarding 'source', it seems difficult to pass the whole source as the parameter.
I suggest to handle *only* 'url' for now.

lib/api.js
```diff
@@ -0,0 +1,91 @@
...
+    else if (query.file) result.file = query.file;
+    else if (query.document) result.document = query.document;
+    if (query.profile) {
+        if (profiles[query.profile])
+            result.profile = profiles[query.profile];
+        else
+            throw new Error('Cannot retrieve profile “' + query.profile + '”');
+    }
+    if (query.validation) result.validation = query.validation;
+    if (query.processDocument) result.processDocument = query.processDocument;
```
deniak: Other options are possible (rectrack, echidna, informative, etc...)

lib/api.js
tripu: Fixed, thanks!

README.md
tripu: That caveat about `file` and `document` isn't new. Those two options don't work for users of our public instance of Specberus, but may come handy in other circumstances, eg we use `file` for testing the rules (and I'm using it now to test the REST API too). Or somebody may want to install Specberus locally and pass along `file`s.

@deniak, I'll modify my tests to test both with `file` and `url`, and maybe exclude the other two from the REST API, OK?

lib/api.js
tripu: You are right, I should be accepting the complete set of parameters.

@deniak, I'm now refactoring the parsing of parameters, because otherwise it'd be duplicated in different places. That common logic will be used then by `lib/validator.js` and `lib/api.js` to parse their parameters. Sounds good?

deniak: @tripu WFM. Don't forget to take the profile `auto` into consideration.

README.md
deniak: `file` and `document` are definitely useful but so far, we know, it's not working properly and these are new features compared to the current pubrules, so :+1: to disabling these options from the REST API for now.

tripu:
@deniak, I think I addressed all comments, and API tests are now more comprehensive.
Take a look, please.

Since the nsp vulnerability is fixed, and this introduces a new interface for Specberus, when merged I'll bump the minor version and publish, and also update the version of Specberus in Echidna, OK?

lib/api.js
```diff
@@ -0,0 +1,122 @@
...
+            options = util.processParams(
+                req.query,
+                undefined,
+                {required: (validate ? ['profile'] : []), forbidden: ['document', 'source']}
+            );
+        }
+        catch (err) {
+            return res.status(400).send(err.toString());
+        }
+        if (validate && 'auto' === options.profile) {
```
deniak: Trying to validate an url with profile `auto` returns the profile only, eg.
`/api/validate?url=http://www.w3.org/TR/2016/WD-charmod-norm-20160407/&profile=auto` gives `WD`

tripu:
@deniak, now `metadata` and `validate` return a JSON object with errors, warnings, metadata, etc. It's documented [on the README](https://github.com/w3c/specberus/blob/tripu/api/README.md#5-rest-api).

/cc @bert-github, who I think is interested in using Specberus programmatically via the REST API.

lib/api.js
deniak: Looks good now

lib/api.js
```diff
@@ -0,0 +1,154 @@
...
+        ;
+        try {
+            options = util.processParams(
+                req.query,
+                undefined,
+                {required: (validate ? ['profile'] : []), forbidden: ['document', 'source']}
+            );
+        }
+        catch (err) {
+            return res.status(400).send(err.toString());
```
deniak: Since the API is expected to return a JSON, it might be better to send a JSON in case of errors too. Same comment for line 145.

deniak:
If the API is used on an url that doesn't exist, the exception isn't caught and the page loads indefinitely.

tripu:
@deniak, back to you.

PR #311, opened 2016-04-06
extractMetadata rectrack

deniak:
`extractMetadata` now detects whether the document is on REC-track or not.
The goal is to pass that information to Echidna so it can also publish document on Note-track.


tripu:
Thank you for simplifying my function for testing metadata! I don't know what I was thinking when I came up with that :)

All good, pending your review of #314. Merge both yourself!

deniak:
@tripu I went ahead and remove the flag `i` from the regex. I don't think it should be case insensitive.

PR #314, opened 2016-04-07
Suggestions to #311

tripu:
Two suggestions:

* Update documentation comment and remove obsolete constants.
* Make the new regex more robust, and make the sample doc a bit tougher.

Feel free to pick and choose, @deniak.

### DIFF
--- a/lib/rules/metadata/rectrack.js
+++ b/lib/rules/metadata/rectrack.js
@@ -7,7 +7,7 @@ exports.name = 'metadata.rectrack';
 exports.check = function(sr, done) {
 
     var $sotd = sr.getSotDSection()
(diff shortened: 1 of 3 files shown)


lib/rules/metadata/rectrack.js
```diff
@@ -7,7 +7,7 @@ exports.name = 'metadata.rectrack';
 exports.check = function(sr, done) {
 
     var $sotd = sr.getSotDSection()
-    ,   expected = /The (group does|groups do) not expect this document to become a W3C Recommendation/;
+    ,   expected = /the\s+(group\s+does|groups\s+do)\s+not\s+expect\s+this\s+document\s+to\s+become\s+a\s+w3c\s+recommendation/i;
```
deniak: Agree with `\s` but I wouldn't put the regex case insensitive.

PR #324, opened 2016-04-15
Usual dependency updates

deniak:


deniak:
Becomes useless with  #327

PR #331, opened 2016-04-21
Handle all the metadata in `extractMetadata`

deniak:
`extractMetadata` takes care of extracting all the metadata needed. There is still an event metadata `shortname` but it's only needed in the exception system. I'll see how to change that in another PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Once the [event `end-all`](#validation-events) is emitted, the metadata should b
 The `options` accepted are equal to those in `validate()`, except that a `profile` is not necessary and will be ignored (finding out the profile is one of the
 goals of this method).
 
-`this.meta` will be an `Object` and may include up to 10 properties:
+`this.meta` will be an `Object` and may include up to 12 properties:
 * `profile`
 * `title`: The (possible) title of the document.
 * `docDate`: The date associated to the document.
@@ -113,6 +113,7 @@ This is an example of the value of `Specberus.meta` after the execution of `Spec
   "editorsDraft": 'http://w3c.github.io/foobar/',
   "delivererIDs": [123, 456],
   "editorIDs": [ 12345 ],
+  "informative": false,
   "rectrack": true,
   "process": 'http://www.w3.org/2015/Process-20150901/' }
 }
--- a/README.md
+++ b/README.md
@@ -94,7 +94,8 @@ goals of this method).
 * `editorsDraft`: URL of the latest editor's draft.
 * `delivererIDs`: ID(s) of the deliverer(s); an `Array` of `Number`s.
 * `editorIDs`: ID(s) of the editor(s) responsible for the document; an `Array` of `Number`s.
-* `rectrack`: Whether the document in on REC track
+* `informative`: Whether the document in informative or not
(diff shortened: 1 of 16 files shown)


lib/validator.js
```diff
@@ -258,6 +257,46 @@ Specberus.prototype.getSotDSection = function () {
...
+    else if (txt.lastIndexOf("latest version", 0) === 0) key = "Latest";
+    else if (/^previous version(?:s)?$/.test(txt)) key = "Previous";
+    else if (/^rescinds this recommendation?$/.test(txt)) key = "Rescinds";
+    if (EDITORS_DRAFT.test(txt) && $dd.find('a')) key = "Editor";
+    if (key) dts[key] = { pos: idx, el: $dt, dd: $dd };
+  });
+  return dts;
+};
+
+Specberus.prototype.getEditorIDs = function () {
```
tripu: Shouldn't the logic of this function be inside `lib/rules/metadata/editor-ids.js` instead?
`lib/validator.js` can store its output locally when it runs, and later serve that value to other rules that need it (eg `lib/rules/echidna/editor-ids.js`).

Two reasons: keep all metadata extraction under `lib/rules/metadata/`, and scan the document in search of editor IDs only once (as it is now, it runs twice, iiuc).

README.md
```diff
@@ -84,86 +84,40 @@ Once the [event `end-all`](#validation-events) is emitted, the metadata should b
 The `options` accepted are equal to those in `validate()`, except that a `profile` is not necessary and will be ignored (finding out the profile is one of the
 goals of this method).
 
-`this.meta` will be an `Object` and may include up to 3 properties: `profile`, `delivererIDs` and `rectrack`.
+`this.meta` will be an `Object` and may include up to 10 properties:
```
tripu: 12?

README.md
```diff
@@ -84,86 +84,40 @@ Once the [event `end-all`](#validation-events) is emitted, the metadata should b
...
+  "docDate": '2016-2-3',
+  "thisVersion": 'http://www.w3.org/TR/2016/WD-foobar-20160203/',
+  "latestVersion": 'http://www.w3.org/TR/foobar/',
+  "previousVersion": 'http://www.w3.org/TR/2015/WD-foobar-20150101/',
+  "editorsDraft": 'http://w3c.github.io/foobar/',
+  "delivererIDs": [123, 456],
+  "editorIDs": [ 12345 ],
+  "rectrack": true,
+  "process": 'http://www.w3.org/2015/Process-20150901/' }
 }
```
tripu: Add sample value for `informative` too?

tripu:
@deniak, this is terrific work!
OK to merge it, if not for the three minor comments inline.

lib/validator.js
deniak: If the logic is moved in a rule that is used in the `metadata` profile, then [Specberus.validate](https://github.com/w3c/specberus/blob/master/lib/validator.js#L74) relies on a preliminary run of the metadata extraction which is not needed in some situations.
Also, `validator.js` is already getting other data such as the SotD or the document date.

tripu: Mmmh I guess you're right. Separation of tasks isn't as clean as we'd like yet.

PR #333, opened 2016-04-22
clean exceptions now that the CSS validator handles calc()

deniak:
@dontcallmedom, since most of the exceptions are for the specs from the DAP WG, do mind reviewing that one?


deniak:
Thanks @dontcallmedom

PR #334, opened 2016-04-25
notice update: announcement the upcoming pubrules switch

deniak:


plehegar:
It looks like we should be able to mandate HTML5 for /TR documents and avoid wondering how to handle the validation for HTML4. Both text/html and application/xhtml+xml are supported in the validator according to Mike.
Looking briefly at past publications in 2015 and 2016, we have CSS2, several WAI doc, EXI, xquery/xslt docs, FileAPI, Web Storage. Imho, we should:
1. figure out why WAI is generating HTML4 and help them move to HTML5. 
2. Tell xquery/xslt and EXI to use text/html with HTML5 syntax or use application/xhtml+xml with XHTML5 syntax.
3. Tell Bert, FileAPI and WebStorage to use HTML5

Philippe

PR #335, opened 2016-04-25
New profile 'auto' using `extractMetadata` to grab the options

### LINKED ISSUE #302: Let users of web page leave profile to "auto"
Once #299 is merged, the web UI of Specberus should change: the profile will be inferred automatically by default, and forced only if the user explicitly chooses that. In the first case, `extractMetadata()` has to be invoked first, and if it goes OK and finds a profile, use it with `validate()` (if something goes wrong, inform the user).

NB: the document should not have to be retrieved twice; instead, keep a snapshot in memory after the first method, and use that for the second one.

deniak:
This PR relies on https://github.com/w3c/specberus/pull/331 to get all the options needed to validate a document.
Fix #302

### DIFF
--- a/public/js/specberus.js
+++ b/public/js/specberus.js
@@ -394,7 +394,16 @@ jQuery.extend({
             var options = $.getQueryParameters();
             setFormParams(options);
             $(".manual").toggle($profile.val() !== 'auto');
-            if (options.url && options.profile) validate(options);
+            if ($profile.val() === "auto") {
+                extractMetadata($url.val());
+            }
+            if (options.url && options.profile) {
+              if (options.profile === "auto") {
+                extractMetadata(options.url);
+              } else {
+                validate(options);
+              }
+            }
         });
     });
 
--- a/public/js/specberus.js
+++ b/public/js/specberus.js
@@ -381,7 +381,7 @@ jQuery.extend({
 
     $(document).ready(function() {
         $.getJSON('data/profiles.json', function(data) {
-            $profile.append($('<option value="auto">Auto-detect</option>'));
+            $profile.append($('<option value="auto" selected="selected">Auto-detect</option>'));
             $.each(data.tracks, function(foo, track) {
                 optgroup = $('<optgroup label="' + track.name + '"></optgroup>');
                 $.each(track.profiles, function(bar, profile) {
--- a/public/js/specberus.js
+++ b/public/js/specberus.js
@@ -394,9 +394,6 @@ jQuery.extend({
             var options = $.getQueryParameters();
             setFormParams(options);
             $(".manual").toggle($profile.val() !== 'auto');
-            if ($profile.val() === "auto") {
-                extractMetadata($url.val());
-            }
             if (options.url && options.profile) {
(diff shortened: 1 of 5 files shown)


public/js/specberus.js
```diff
@@ -192,35 +199,62 @@ jQuery.extend({
...
-        ,   validation = $validation.val()
-        ,   noRecTrack = $noRecTrack.is(":checked") || false
-        ,   informativeOnly = $informativeOnly.is(":checked") || false
-        ,   echidnaReady = $echidnaReady.is(":checked") || false
-        ,   patentPolicy = $patentPolicy.find('label.active').attr('id')
-        ,   processDocument = $processDocument.find('label.active').attr('id')
-        ;
-        if (!url) showError("Missing URL parameter.");
-        if (!profile) showError("Missing profile parameter.");
-        if (echidnaReady) profile += '-Echidna';
```
tripu: If in the UI you tick the checkbox `#echidnaReady`, the `Auto-detect` profile gets disabled, and you can't select it again.
I think it's because in this line `-Echidna` is appended to the profile that is selected, and the only profile that exists both with and without `-Echidna` is `WD`.
Perhaps we should make an exception if the selected profile is `auto` (or have `auto-Echidna` too).

public/js/specberus.js
```diff
@@ -192,35 +199,62 @@ jQuery.extend({
...
+      $noRecTrack.prop('checked', !data.rectrack);
+      $informativeOnly.prop('checked', data.informative);
+      $validation.val('simple-validation');
+      var options = {
+                        "url"             : data.url
+                      , "profile"         : data.profile
+                      , "validation"      : 'simple-validation'
+                      , "noRecTrack"      : !data.rectrack || false
+                      , "informativeOnly" : data.informative || false
+                      , "echidnaReady"    : false
```
tripu: Why `echidnaReady = false` here?

tripu:
What do you think if we have `auto` selected by default, and hide all other options, @deniak?

Simplest UI, less intimidating, less cognitive dissonance. Those users who *know* what they need will find their way in the drop-down box anyway&hellip;

public/js/specberus.js
deniak: I think echidna is a very special case of Specberus. When you select that option, you can't expect specberus to check a FPWD, CR, PR or REC.
Furthermore, when the auto profile is selected, Specberus will ignore the options (informative, rec-track, profile, etc) so it can be confusing if the option echidna is treated differently.

public/js/specberus.js
deniak: Because, I think echidna is a special use case. This is not something that can be extracted from the document so the "default" run the echidna rules. The options can still be changed after the 1st run if necessary.

deniak:
@tripu the `auto` profile is selected by default. The options appear only if you select a different profile.

tripu:
@deniak, if you refresh the page (without cache), `$('#profile').val()` is `null` and all the options are visible.

Feel free to merge yourself if I'm wrong about that, or if you change it :)

PR #336, opened 2016-04-25
'true' and 'false' from QS should be treated as booleans and not strings

deniak:


public/js/specberus.js
```diff
@@ -299,7 +299,7 @@ jQuery.extend({
 
     function setFormParams(options) {
         // Option "echidnaReady" processed first, as it may restrict the list of enabled profiles.
-        if (options.echidnaReady === "true") $echidnaReady.prop('checked', true);
+        if (options.echidnaReady === true) $echidnaReady.prop('checked', true);
```
tripu: I suggest `$echidnaReady.prop('checked', options.echidnaReady);` instead.
Same in the other 2 lines below.

tripu:
LGTM, @deniak.
I only have a suggestion to make those lines more concise, but good to merge!

PR #339, opened 2016-04-27
new endpoint to replace getdocrdf

deniak:


deniak:
The build is failing because of `[engine.io-client](https://www.npmjs.com/package/engine.io-client)` not being up-to-date. It will be fixed after a new version is published.

PR #345, opened 2016-05-02
fix extractHeaders on the same day publication check

deniak:
That PR fixes the same day publication check as `extractHeaders` is now defined in `Specberus`.


deniak:
Fix #348

PR #353, opened 2016-05-06
Make ToC detection more accurate

### LINKED ISSUE #351: False negative for document containing nav element.
The document at [1] contains a ``nav`` element with a child ``h2`` element with text content "Table of Contents"; however, the new pubrules checker 1.5.3 reports an error:

✘ fail	headers.h2-toc	
error There is no table of contents inside a navigation element (``<nav id="toc"> … <h2>Table of Contents</h2> … </nav>``).

The document contains ``<nav id="toc"><h2 class="introductory" id="table-of-contents">Table of Contents</h2>...``. Perhaps it doesn't like the presence of ``@class`` and ``@id`` on ``h2``?

[1] https://rawgit.com/w3c/ttml2/ttml2-WD-201605XX/spec/ttml2.html

tripu:
More than one `<h2>` inside `<nav id="toc">` is now permitted, as long as exactly one matches `/^table\s+of\s+contents$/i`.

Fixes #351.

### DIFF
--- a/lib/l10n-en_GB.js
+++ b/lib/l10n-en_GB.js
@@ -21,10 +21,11 @@ exports.messages = {
 ,   "headers.h2-status.no-h2":  "No status+date h2 element."
 ,   "headers.h2-status.bad-h2": "Incorrect status h2 header."
     // headers/h2-toc
-,   'headers.h2-toc.not-found':  'There is no table of contents inside a navigation element (<code>&lt;nav id="toc"&gt; &hellip; &lt;h2&gt;Table of Contents&lt;/h2&gt; &hellip; &lt;/nav&gt;</code>).'
-,   'headers.h2-toc.too-many':   'Found more than one navigation elements; possibly containing the table of contents. There should be just one <code>&lt;nav id="toc"&gt;</code>.'
-,   'headers.h2-toc.not-html5':  'Support for old HTML doctypes is deprecated and will be abandoned soon; please upgrade to HTML5. The expected navigation element is: <code>&lt;nav id="toc"&gt;</code>.'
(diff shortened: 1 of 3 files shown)


deniak:
If a document is using more than one `div#nav > h2`, the PR will report an error whatever the text nodes are. The rule can be factorised to have the same behaviour for both `nav#toc > h2` and `div#toc > h2` with the additional warning for the latter.

tripu:
Thanks, @deniak. I think I fixed that now.

deniak:
@tripu LGTM although I think it can be simplified to avoid repetition:

```javascript
    const EXPECTED_HEADING = /^table\s+of\s+contents$/i
    ,   tocDiv = sr.$('div#toc > h2')
    ;

    var toc = sr.$('nav#toc > h2');

    if(tocDiv.length > 0) {
        if(toc.length > 0)
            sr.error(exports.name, 'mixed');
        else {
            sr.warning(exports.name, 'not-html5');
            toc = tocDiv;
        }
    }

    if (toc && toc.length > 0) {
        var matches = 0;
        for (var i = 0; i < toc.length; i ++)
            if (EXPECTED_HEADING.test(sr.$(toc[i]).text()))
                matches ++;
        if (matches > 1)
            sr.error(exports.name, 'too-many');
        else if (0 === matches)
            sr.error(exports.name, 'not-found');
    }

    done();
```

tripu:
I think that's not equivalent, @deniak:

* If `0 === tocDiv.length && 0 === toc.length`, the rule reaches `done()` and is OK (wrongly).
* If both styles are found, but the regex is not satisfied, two errors (`mixed` *and* `not-found`) will be returned, which will seem contradictory to the user.

deniak:
Ok, I see your point. I was trying to remove the duplicate `sr.error(exports.name, 'not-found');` but it doesn't catch the use cases you mentioned.

PR #355, opened 2016-05-07
New navigation structure, UI and pubrules documentation

tripu:
Open questions:

## 1. Navigation

This is the current sitemap:
* `/` (will be deployed at `https://www.w3.org/pubrules/`)
  * `doc/`: documentation
    * `rules/`: publication rules
    * `transitions/`: transition rules
    * Other rules and process documents will go here too
  * `check/`: checker (the old Specberus)
  * `about/`: &ldquo;about&rdquo; page
  * `sitemap/`: sitemap
  * `help/`: help, tutorials, FAQ, API reference, etc.

**Does it look right?**

## 2. URIs of documentation

Currently, *pubrules* are found here:

* `/doc/rules/?profile=WD`
* `/doc/rules/?profile=REC`
* etc

and other documents, like *transition rules*, will be at:

* `/doc/transition/docstatus=fpwg-note-tr&notehistory=closure`
* `/doc/transition/docstatus=pr-tr&skipcr=on`
* etc

ie, using **query strings**.

The alternative would be using **directories in the path**, like this:

* `/doc/rules/WD`
* `/doc/rules/REC`
* `/doc/transition/fpwg-note-tr/closure/`
* `/doc/transition/pr-tr/skipcr/`
* etc

**Which one will work best (for URI manipulation, caching, etc)?**

## 3. `lib/rules.json`

The file as it is has a few flaws:

* Using object keys in some places of the structure where arrays would be easier to handle.
* Missing sections (eg, no. 4 missing in most profiles).
* Unordered sections and profiles.
* Some keys are `undefined` (eg, section no. 6 of all profiles).

`lib/rules-wrapper.js` now corrects some of these errors, but&hellip;

**&hellip;can we correct the generation script, and regenerate `rules.json` to address these issues?**

## 4. Other rules

Pending: extract transition rules and other structured docs from original pubrules viewer, and start using them here. Ideas welcome.

## 5. Avoid redundancy; simplify JSON

For *pubrules* and any other rules we store/show: can we do something clever to avoid the huge amount of redundancy that there is now?

eg, many rules are identical for different profiles, yet we store them for each of them. Other rules vary only slightly per profile; we could make them *parametric* and store the rule as a template, plus the different bits&hellip;

## 6. Pending about pubrules display

* Offset height of the header overlay when linking to a fragment ID, so that the section is not hidden underneath it.
* Improve presentation of highlighted heading or rule.
* Style class `rfc2119` and others.
* Add an index of sections for better navigation.
* Let the user collapse/expand sections (?).


deniak:
## 1. Navigation

> This is the current sitemap:
> * `/` (will be deployed at `https://www.w3.org/pubrules/`)
>   * `doc/`: documentation
>     * `rules/`: publication rules
>     * `transitions/`: transition rules
>     * Other rules and process documents will go here too
>   * `check/`: checker (the old Specberus)
>   * `about/`: &ldquo;about&rdquo; page
>   * `sitemap/`: sitemap
>   * `help/`: help, tutorials, FAQ, API reference, etc.
> 
> **Does it look right?**

Can we have the form on the homepage (merge `/` and `check/`)and some kind of navigation next to the form? So we keep the same structure as the old pubrules.
Otherwise the sitemap looks good.

## 2. URIs of documentation

> Currently, *pubrules* are found here:
> 
> * `/doc/rules/?profile=WD`
> * `/doc/rules/?profile=REC`
> * etc
> 
> and other documents, like *transition rules*, will be at:
> 
> * `/doc/transition/docstatus=fpwg-note-tr&notehistory=closure`
> * `/doc/transition/docstatus=pr-tr&skipcr=on`
> * etc
> 
> ie, using **query strings**.
> 
> The alternative would be using **directories in the path**, like this:
> 
> * `/doc/rules/WD`
> * `/doc/rules/REC`
> * `/doc/transition/fpwg-note-tr/closure/`
> * `/doc/transition/pr-tr/skipcr/`
> * etc
> 
> **Which one will work best (for URI manipulation, caching, etc)?**

:+1:  for QS params

## 3. `lib/rules.json`

> The file as it is has a few flaws:
> 
> * Using object keys in some places of the structure where arrays would be easier to handle.
> * Missing sections (eg, no. 4 missing in most profiles).
> * Unordered sections and profiles.
> * Some keys are `undefined` (eg, section no. 6 of all profiles).
> 
> `lib/rules-wrapper.js` now corrects some of these errors, but&hellip;
> 
> **&hellip;can we correct the generation script, and regenerate `rules.json` to address these issues?**

`rules.json` can be updated as needed. I'll address the issues

## 4. Other rules

> Pending: extract transition rules and other structured docs from original pubrules viewer, and start using them here. Ideas welcome.

We can probably use another json file like we are doing for `rules.json`. I'll try to provide this.

## 5. Avoid redundancy; simplify JSON

> For *pubrules* and any other rules we store/show: can we do something clever to avoid the huge amount of redundancy that there is now?
> 
> eg, many rules are identical for different profiles, yet we store them for each of them. Other rules vary only slightly per profile; we could make them *parametric* and store the rule as a template, plus the different bits&hellip;

We should use the same mechanism as we are doing for `l10n-en_GB.js`. Template messages.

tripu:
@deniak, this is ready, barring more tests and checking a couple of implementation details with you. Let's chat about it.

lib/l10n.js
```diff
@@ -1,29 +1,74 @@
...
+    const result = {};
+    var name;
+    if ('string' === typeof rule)
+        name = rule;
+    else
+        name = rule.name;
+    if (!lang)
+        throw new Error('l10n.message() invoked before a locale is defined; call l10n.setLanguage() first');
+    else if (!rulesWrapper.hasOwnProperty(profile))
+        throw new Error(`l10n.message(): unknown profile code “${profile}”`);
```
deniak: Checking a document with the wrong profile makes the app crash. For instance, a WD with the profile Rec gives:
```
Error: l10n.message(): unknown profile code “Rec”
    at Object.exports.message (.../specberus/lib/l10n.js:51:15)
    at null.<anonymous> (.../specberus/app.js:96:37)
    at emitTwo (events.js:105:20)
    at emit (events.js:185:7)
    at Specberus.error (.../specberus/lib/validator.js:138:23)
    at err (.../specberus/lib/rules/headers/dl.js:28:16)
    at Object.exports.check (.../specberus/lib/rules/headers/dl.js:94:14)
    at .../specberus/lib/validator.js:107:18
    at Array.forEach (native)
    at doValidation (.../specberus/lib/validator.js:103:23)
```

These errors should be caught so the app doesn't crash.

deniak:
For all the bootstrap assets, it's better to use the ones in https://www.w3.org/scripts/ to avoid duplications.

views/index.handlebars
```diff
@@ -0,0 +1,155 @@
...
+                <option value disabled="disabled">Select an option</option>
+              </select>
+            </div>
+            <div id="check1" class="col-sm-2 col-md-1 col-lg-1">
+              <button type="submit" class="btn btn-primary pull-right">Check</button>
+            </div>
+          </div>
+
+          <div class="form-group manual">
+            <label class="col-xs-3 col-md-1 col-lg-1 control-label"><abbr title="Patent Policy">PP</abbr></label>
```
deniak: I suggest `s/PP/Patent Policy/` and remove `<abbr>` to have a clear interface. PP is not an acronym people are familiar with outside the team.

views/index.handlebars
```diff
@@ -0,0 +1,155 @@
...
+  only that one is considered authoritative.<br>
+  Contribute and submit issues <a href="https://github.com/w3c/specberus">on GitHub</a>; send comments to
+  <a href="mailto:public-pubrules-comments@w3.org"><code>public-pubrules-comments@w3.org</code></a>.
+</div>
+
+<div class="row">
+  <div class="col-md-12">
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        <h3 class="panel-title">Test a document against the <a href="/doc">publication rules</a></h3>
```
deniak: Add a link to the transitions guide: http://www.w3.org/Guide/transitions2015

views/doc/index.handlebars
```diff
@@ -0,0 +1,18 @@
...
+
+    <h3 id="publication"><a href="#publication">Publication rules</a></h3>
+    <ul>
+      {{#each profiles}}
+        <li><a href="/doc/rules?profile={{this.abbr}}">{{this.name}} ({{this.abbr}})</a></li>
+      {{/each}}
+    </ul>
+
+    <h3 id="transitions"><a href="#transitions">Transitions</a></h3>
+    <p><a href="/doc/transitions/">Coming soon</a></p>
```
deniak: Point to the current transitions guide: http://www.w3.org/Guide/transitions2015

views/doc/index.handlebars
```diff
@@ -0,0 +1,18 @@
...
+  <div class="col-md-12">
+    {{{nav}}}
+    <p><a href="../">&uarr; up to Specberus home</a></p>
+    <h2>Specberus documentation</h2>
+
+    <h3 id="publication"><a href="#publication">Publication rules</a></h3>
+    <ul>
+      {{#each profiles}}
+        <li><a href="/doc/rules?profile={{this.abbr}}">{{this.name}} ({{this.abbr}})</a></li>
+      {{/each}}
```
deniak: Seems like we are missing the submissions doc. We don't receive many submissions so we can keep that for later. I created a new ticket so we don't forget it: #369

views/index.handlebars
```diff
@@ -0,0 +1,155 @@
...
+      </div>
+      <div class="panel-body">
+        <form id="options" class="form-horizontal manual">
+
+          <div class="form-group">
+            <label for="url" class="col-xs-3 col-sm-1 col-md-1 col-lg-1 control-label">URL</label>
+            <div id="urlContainer" class="col-xs-9 col-sm-4 col-md-5 col-lg-5">
+              <input type="url" class="form-control" id="url" required autofocus="autofocus" value="https://">
```
deniak: use `placeholder="https://www.w3.org/TR/2016/WD-foobar-YYYYMMDD/` instead of `value="https://"` ?
This way people will benefit from client side validation if they try to submit the form without adding a url.

views/index.handlebars
```diff
@@ -0,0 +1,155 @@
...
+      <div class="progress-bar" role="progressbar" aria-valuenow="0" aria-valuemin="0" aria-valuemax="0" style="width: 0">
+        <span class="">x/y done</span>
+      </div>
+    </div>
+  </div>
+</div>
+
+<div class="row">
+  <div id="alert" class="col-md-12">
+    <div class="alert alert-danger alert-dismissable hide">
```
deniak: New alerts shouldn't be hidden.
Eg. validating 'http://localhost:1234' produces `[EXCEPTION] connect ECONNREFUSED 127.0.0.1:1234`. The DOM is updated but because of the class `hide`, the alert doesn't appear on the UI.

tripu:
@deniak:

**1. Navigation**

> Can we have the form on the homepage (merge `/` and `check/`)and some kind of navigation next to the form? So we keep the same structure as the old pubrules. Otherwise the sitemap looks good.

Done.

**2. URIs of documentation**

> :+1:  for QS params

OK!

**3. `lib/rules.json`**

> `rules.json` can be updated as needed. I'll address the issues

Great!

If you do &ldquo;fix&rdquo; the file, bear in mind that metadata of rules rely on the oddities of the current [`lib/rules.json`](https://github.com/w3c/specberus/blob/tripu/ui-tweaks/lib/rules.json):

* The current key `'undefined'` is used as such ([example](https://github.com/w3c/specberus/blob/tripu/ui-tweaks/lib/rules/headers/ol-toc.js#L7-L8), [example](https://github.com/w3c/specberus/blob/tripu/ui-tweaks/lib/rules/structure/h2.js#L15-L16)).  
Find them with `$ grep -R "'undefined'" lib/rules/*`.
* Not all rules we're applying exist in human-readable form in `rules.json` ([example](https://github.com/w3c/specberus/blob/tripu/ui-tweaks/lib/rules/echidna/editor-ids.js#L4-L5), [four examples](https://github.com/w3c/specberus/blob/tripu/ui-tweaks/lib/rules/structure/display-only.js#L9-L12)). We'll have to add those.

Also, [`lib/rules-wrapper.js`](https://github.com/w3c/specberus/blob/tripu/ui-tweaks/lib/rules-wrapper.js) *may* become obsolete if you transform `rules.json` enough.

**4. Other rules**

> We can probably use another json file like we are doing for `rules.json`. I'll try to provide this.

(We saw that we don't need to import other XSLT rules from the old checker; at least for now.)

**5. Avoid redundancy; simplify JSON**

> We should use the same mechanism as we are doing for `l10n-en_GB.js`. Template messages.

I ended up with a (hopefully) better mechanism: embedding metadata about the rule in the rule itself. [`lib/l10n-selectors.js` is not necessary any more](https://github.com/w3c/specberus/pull/355/commits/e1f3913c464a952407be61e73cf9b4941cc6b528#diff-092a9b2ad6fd16270685803cef0c4ad5).

tripu:
@deniak, I'm addressing your comments; will commit my changes soon.

views/doc/index.handlebars
tripu: OK!

views/index.handlebars
tripu: I don't understand thigs, @deniak. Client-side validation?

deniak:
Amazing job @tripu !

PR #360, opened 2016-05-18
detect the feedback due date following what pubrules' current rule

deniak:


lib/rules/metadata/profile.js
```diff
@@ -41,7 +41,19 @@ exports.check = function(sr, done) {
...
     });
     if (id) {
-        done({profile: id});
+        if (id === "CR" || id === "PR" || id === "PER") {
+            var dates = sr.getFeedbackDueDate()
+            ,   res   = dates[0]
+            ;
+            if (dates.length === 0 || !res) return done({profile: id});
+            if (dates.length > 1)
+                res = new Date(Math.min.apply(null,dates));
```
tripu: (In ES6, `Math.min(...dates)` is even shorter :¬)

PR #361, opened 2016-05-18
new rule to check consistency between pp link and WG homepage

### LINKED ISSUE #316: Make sure PP link matches the right WG
The current `pp` rule doesn't check that the pp link is the right one

deniak:
Fix #316


deniak:
@tripu that rule is sending warnings instead of errors because of the tests. It relies on the W3C API and I don't want to hardcode a key in GH. I'm not sure how to properly handle this. In any case, we cannot fully test the rule because of the API key.

deniak:
Using a config file to pass the API key doesn't seem to be a good choice. It will make it harder to use by Echidna so I changed the PR to get the key from an environment variable `API_KEY`.
My comment about the warnings and travis still applies.

PR #378, opened 2016-05-26
Errors and rule descriptions for echidna profiles

### LINKED ISSUE #377: Error messages displayed if the Echidna option is checked
This happens because the profile is not a standard one:
`l10n.message(): unknown profile code “WD-Echidna”`

deniak:
Since the echidna profiles are not standard, use the related profile to display error messages
Fix #377


tripu:
Oops. Well spotted!

PR #384, opened 2016-05-30
Use env var “BASE_URI” for links/routes instead of assuming root dir

tripu:
And update the documentation.


deniak:
Can you make BASE_URI optional? If it's not specified, use `/`.
Also, we might be using the same instance from different hosts so it would be better to *not* specify the host in the variable

tripu:
Ready again, @deniak!

PR #387, opened 2016-05-30
fix profiles.json path if location.pathname='/'

deniak:


tripu:
That slash is necessary if the URL doesn't end with a slash, and doesn't hurt if not.

Am I wrong?

PR #388, opened 2016-05-30
update rule description for normative representation: support HTML5 only

deniak:
@plehegar this PR updates the rule description about the normative representation now that we support HTML5 documents only


plehegar:
I don't think we need a fallback alternative to SVG 1.1 anymore and I don't think we enforce the Math 2.0 alternative. I suggest the following change:
[[
with the following limitations: <ul> <li>Inline markup for SVG 1.1 or MathML 2.0 is permitted but only with a (fallback) alternative.
]]
should read
[[
. Inline markup for MathML is permitted but should use a (fallback) alternative.
]]

deniak:
@plh done

plehegar:
:+1:

PR #389, opened 2016-05-30
It should now be called pubrules

deniak:


tripu:
@deniak, I committed a tiny correction before merging (that message should never be displayed anyway, but just in case).

PR #407, opened 2016-06-03
Multiple latest version URIs

### LINKED ISSUE #402: If there are multiple 'Latest Version' links, take the first one
See http://www.w3.org/TR/2016/WD-html51-20160602/

deniak:
If multiple latest version URIs are mentioned in the document (eg. http://www.w3.org/TR/2016/WD-html51-20160602/), Specberus will take the first one. Fix #402 

That PR also updates the regex used to extract the editor's draft to catch more ED.


tripu:
@deniak, OK about the change. But, shouldn't we make the new criteria explicit in the wording of the rules, [as @plehegar seemed to suggest](https://github.com/w3c/specberus/issues/402#issuecomment-223244051)?

deniak:
@tripu I updated the rule description

PR #420, opened 2016-06-10
Look for a GitHub repository, and qualify the ML/subscribe checks accordingly

### LINKED ISSUE #408: "Mailing List" check doesn't account for groups using GitHub Issues
The CSSWG, CSS Houdini TF, and FXTF are all using GitHub Issues as their primary communication mechanism.  We're no longer recommending using www-style for general comment; it's reserved for publication announcements, pings for review, and the occasional general-interest question.

As such, the current CSSWG boilerplate omits explicit mention of sending email to www-style, and instead links to the Issues page for our repo. It does include a link to the mail archive, however.  (See the second paragraph of <https://drafts.csswg.org/css-values/#status> for an example.)

I'm currently failing Specberus due to this.  It would be nice to not have to lie about our preferred communication medium just to pass.

tabatkins:
Fixes #408.

### DIFF
--- a/lib/rules/sotd/mailing-list.js
+++ b/lib/rules/sotd/mailing-list.js
@@ -20,6 +20,7 @@ exports.check = function (sr, done) {
     var foundML = false
     ,   foundSub = false
     ,   foundArch = false
+    ,   foundRepo = false
     ;
     $sotd.find("a[href]").each(function () {
         var $a = sr.$(this)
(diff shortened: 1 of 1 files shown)


tabatkins:
@deniak @plehegar This looks like a spurious CI failure.

lib/rules/sotd/mailing-list.js
```diff
@@ -39,9 +40,15 @@ exports.check = function (sr, done) {
             foundArch = true;
             return;
         }
+        if (/^https:\/\/github.com\/[\w-]+\/[\w-]+/.test(href)) {
```
deniak: For that rule, it might be safer to look for the issues of the repo (`/^https:\/\/github.com\/[\w-]+\/[\w-]+/issues`). There could be other GH links in the SotD for different purposes.

tabatkins:
@deniak Ok, went ahead and checked for the issues link specifically. I'm already including a direct link in the CSS boilerplate, so that definitely won't block my publication. ^_^

PR #450, opened 2016-06-24
Fix bug in rule “headers.dl”: shortname change

### LINKED ISSUE #439: Bug in rule “headers.dl”: shortname change
When there's a shortname change, the shortlink doesn't exist yet and the rule reports 2 errors:
* Short names differ between This and Previous Versions.
* Could not retrieve URI for previous version or for latest version.

Found [while checking `https://www.w3.org/TR/2016/WD-css-scroll-snap-1-20160623/`](https://www.w3.org/pubrules/?url=https%3A%2F%2Fwww.w3.org%2FTR%2F2016%2FWD-css-scroll-snap-1-20160623%2F&profile=WD&validation=simple-validation&noRecTrack=false&informativeOnly=false&echidnaReady=false&patentPolicy=pp2004).

tripu:
Fix #439.

&hellip;or so I think.

@deniak, relevant changes are starting on [line ~120](https://github.com/w3c/specberus/pull/450/files#diff-09f641a52baf89d6d1785acc66dafe97L120).

(I also wrote &ldquo;shortname&rdquo; everywhere for consistency, and fixed indentation in a couple places.)


lib/rules/headers/dl.js
```diff
@@ -117,7 +119,19 @@ exports.check = function (sr, done) {
         if (matches) {
             sn = matches[1];
             previousURI = $linkPrev.text();
-            if (sn !== shortname) sr.error(self, "this-previous-shortname");
+            if (sn !== shortname) {
+              fetch(previousURI).then(function(body) {
```
deniak: @tripu if the shortname differs (`sn !== shortname`), you should follow the latest version link, and not the previous version, then if it gives a 404, send a warning instead of an error.
```
<denis> if you want to do it the easy way, you can do that
<denis> - compare shortname from this version and previous version
<denis> - if they are different, follow the shortlink and if it's a 404, there's good chance it's shortname change
<tripu> ...in which case we show a warning
<tripu> (not an error)
<tripu> ?
<denis> yes
<tripu> ok
```

tripu:
I changed that, @deniak. Is it right now?

PR #453, opened 2016-07-05
SVG errors reported by the CSS validator are treated as exceptions

deniak:
@tabatkins @ylafon, that PR makes specberus ignore the SVG errors reported by the CSS validator. 
I did it that way because I expect it will require a lot of work to implement all the SVG properties in the validator. Here's the list of properties that will be skipped:
* `fill`
* `fill-opacity`
* `text-anchor`
* `stroke`
* `stroke-width`
* `stroke-dasharray`

This is a non-exhaustive list but I believe it'll enough for the specs generated by bikeshed.
@ylafon I assigned you as the reviewer to get your input on how we are going to handle these errors.

### DIFF
--- a/lib/exceptions.js
+++ b/lib/exceptions.js
@@ -35,7 +35,9 @@ Exceptions.prototype.has = function (shortname, rule, key, extra) {
             var exception = set[i][y];
 
             if (rule === exception.rule &&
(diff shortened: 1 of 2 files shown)


deniak:
After talking to @ylafon, there's a plan to update the CSS validator but in the meantime, we can ignore these errors.

PR #458, opened 2016-07-11
Fix missing error message

### LINKED ISSUE #386: Bug: missing error message
Specberus sometimes returns confusing error message `Cannot read property 'text' of undefined`.

Found [while checking `http://www.w3.org/TR/2016/WD-vocab-ssn-20160531/`](https://www.w3.org/pubrules/?url=http%3A%2F%2Fwww.w3.org%2FTR%2F2016%2FWD-vocab-ssn-20160531%2F&profile=FPWD&validation=simple-validation&noRecTrack=false&informativeOnly=false&echidnaReady=false&patentPolicy=pp2004).

tripu:
Fix #386.

When the selector `rulesWrapper[profile][rule.section][rule.rule].text` can't be resolved, the error is captured so that it doesn't mask the real issue. Also, a link is provided to file a GH issue to alert maintainers.

@deniak, while testing this, I've discovered that this messages are missing from `lib/rules.json`:

* `[WD][front-matter][hrAfterCopyright]`
* `[FPWD][document-status][changesList]`

Shall I add them, copying from another profile? Or do they need changes in the wording?


deniak:
> * `[WD][front-matter][hrAfterCopyright]`

That one is indeed missing. Can you please add it?

> * `[FPWD][document-status][changesList]`

The rule description refers to changes from the previous draft. Since it's a FPWD, there's no previous version so that rule is not part of the FPWD profile.

tripu:
@deniak, see the 3 most recent commits.

PR #464, opened 2016-07-13
Omnibus PR

tripu:
* Improve `viewport` meta tag.
* Encode also the URL when submitting a GH issue. As it was before, a URL containing ampersands (which happens often when using spec-generator on labs.w3.org) was interpreted by GH as containing parameters *for it*, thus truncating the rest of the markdown.


deniak:
@tripu let me know when that PR is ready for review

tripu:
@deniak, ready!

PR #469, opened 2016-07-18
Prevent redundant run; monitor socket status

### LINKED ISSUE #149: Run button needs to be replaced with Cancel button as soon as it's running
We need to maintain state there, because if you double-click on the Run button (which is guaranteed to happen whenever the UI is responding too slowly), you get the process started multiple times in parallel.
lol

### LINKED ISSUE #399: Warn the user if the socket is mute for some time
Scenario: a user opens the web page (or has had it open for hours), but the server just crashed or had a networking issue. The user will try to validate documents, but there will be no feedback (the button will seem to be not working).

Solution: if there's no heartbeat from the server for the last *n* s, show an error to the user, and prompt them to wait and/or reload the page.

tripu:
* Whole form is disabled (greyed-out) and text of button changes (from &ldquo;check&rdquo; to &ldquo;wait&hellip;&rdquo;) if there's a validation running. Prevents against clicking more than once. Fixes #149.
* When the socket disconnects, reload the page (fixes #399).
* When there are query parameters and validation occurs automatically, do so by actually submitting the form (instead of just sending stuff through the socket). Makes it consistent with the changes above (form is disabled), and hopefully will solve situations where validation isn't triggered automatically now&hellip; (not sure; I'll have to keep on testing that).
* Progress bar returns to 0 once validation finishes (prevents that effect when the progress bar appears to go backwards).

### DIFF
--- a/public/js/specberus.js
+++ b/public/js/specberus.js
@@ -72,11 +72,11 @@ jQuery.extend({
     const toggleForm = (bool) => {
         if (bool) {
             $('form').css('opacity', 1);
-            $('form input').removeClass('disabled').removeAttr('disabled');
+            $('form input, form select, form label').removeClass('disabled').removeAttr('disabled');
             $('button[type=submit]').text('Check');
         } else {
-            $('form').css('opacity', 0.25);
-            $('form input').addClass('disabled').attr('disabled', 'disabled');
+            $('form').css('opacity', 0.333);
+            $('form input, form select, form label').addClass('disabled').attr('disabled', 'disabled');
             $('button[type=submit]').text('Wait…');
         }
     };
--- a/public/js/specberus.js
+++ b/public/js/specberus.js
@@ -186,8 +186,6 @@ jQuery.extend({
     });
     socket.on("finished", function () {
         console.log('Finished.');
-        toggleForm(true);
-        running = false;
         showResults();
     });
     socket.on("finishedExtraction", function (data) {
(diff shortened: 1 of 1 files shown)


public/js/specberus.js
```diff
@@ -54,13 +54,33 @@ jQuery.extend({
...
     });
 
+    const toggleForm = (bool) => {
+        if (bool) {
+            $('form').css('opacity', 1);
+            $('form input').removeClass('disabled').removeAttr('disabled');
+            $('button[type=submit]').text('Check');
+        } else {
+            $('form').css('opacity', 0.25);
+            $('form input').addClass('disabled').attr('disabled', 'disabled');
```
deniak: While specberus is running, it is still possible to update the form. `<label>`s and `<select>` should be disabled as well.

deniak:
If the socket hang up after a run, the form is still grayed out and it's not possible to resubmit a request. Any chance you can trigger `toggleForm(true)` in this case?
specberus-socket-hangup

tripu:
@deniak, if the socket disconnects, [the page should reload](https://github.com/w3c/specberus/pull/469/files#diff-25026d65d1ca3b34938e9cc4fc7e5c01R64).

How did you get to that point in the screenshot? I can't reproduce it by simply stopping the server.

deniak:
I ended up in that state trying to run specberus multiple time in a row with different options. Unfortunately that's hard to reproduce but I saw the exception "socket hang up" was thrown in the logs.

deniak:
@tripu I think calling `toggleForm` in https://github.com/w3c/specberus/blob/master/public/js/specberus.js#L134 should fix the issue

tripu:
@deniak, that last commit achieves the same result (in fact fixes corner cases when there are exceptions).

I can't reproduce that situation on your screenshot, but if that happens, the form should be enabled again now.

PR #480, opened 2016-08-01
Prevent Specberus from reloading the page automatically when connection to the server is lost

### LINKED ISSUE #479: Specberus reload the page by itself after a few minutes
How to reproduce:
- load https://www.w3.org/pubrules/?url=https%3A%2F%2Fwww.w3.org%2FTR%2F2016%2FCR-mixed-content-20160802%2F&profile=CR&validation=simple-validation&noRecTrack=false&informativeOnly=false&echidnaReady=false&patentPolicy=pp2004 
- wait a min or two and specberus will run again

tripu:
@deniak, this is my first take, but I'm not sure. Let me know your thoughts.

When the socket disconnects, a modal appears, informing the user of what just happened. Users can either dismiss and continue examining the results of validation, or reload the page if they finished reading the results already.

The modal won't bother the user a second time (not unless they reload the page and validate a document again).

I like this approach. The problem is, if connection through the socket is interrupted quite often (as it seems; I don't know why), it's a matter of (little) time that users will see this pop-up. I fear it will be annoying and interrupt the user while they're reading the results.

I may not be listening to the correct socket event, too.

Fix #479.

### DIFF
--- a/public/js/specberus.js
+++ b/public/js/specberus.js
@@ -64,7 +64,7 @@ jQuery.extend({
         socket.on('disconnect', () => {
             socket.close();
             toggleForm(false);
-            window.location.href = window.location.href;
+            $('#offline').modal();
         });
 
     });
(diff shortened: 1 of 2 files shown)


deniak:
The socket was fixed directly from the process manager (https://github.com/w3c/specberus/issues/479#issuecomment-238552329). It shouldn't disconnect anymore.

deniak:
Thinking more about it, that's better than reloading the page if the socket is disconnected

PR #486, opened 2016-08-11
don't handle socket disconnection so it doesn't interfere when a user leave pubrules

deniak:
@michael-n-cooper tried to submit a bug report but noticed clicking on the link was causing the page to reload.

### DIFF
--- a/public/js/specberus.js
+++ b/public/js/specberus.js
@@ -58,13 +58,6 @@ jQuery.extend({
     // handshake
     socket.on("handshake", function (data) {
         console.log(`Handshake; using version “${data.version}”.`);
-
-        socket.on('disconnect', function () {
-            socket.close();
-            toggleForm(false);
-            $('#offline').modal();
-        });
-
     });
 
     const toggleForm = function (bool) {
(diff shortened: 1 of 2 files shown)


tripu:
@deniak, I can't reproduce the bug.

@michael-n-cooper, clicking *any* of the &ldquo;Report a bug&rdquo; links always reloads Pubrules (instead of taking you to GitHub to file the issue)?
On what browser does that happen?

deniak:
@tripu It's working fine on chromium but not on firefox 48

tripu:
@deniak. Do *you* reproduce it in FF 48?
I still can't reproduce the issue in FF 48 on Ubuntu 16.04&hellip;

deniak:
yes I can reproduce the bug with FF 48, but don't have the problem in chomium

deniak:
The current master branch (not deployed yet) shows a modal dialog asking if you want to reload or remain on the page if the socket gets disconnected.
On FF48, if someone clicks on 'Report a bug', that modal dialog will appear quickly and then the user will be redirected to github.
Discussing this with Antonio, we decided to keep that behavior but don't show the modal if a user clicks on that link.

PR #503, opened 2016-09-16
Enable errors and warnings from metadata extraction; alert if process document is not found

### LINKED ISSUE #501: Check that the process document is right
&hellip;or at least, that it's not `undefined`.

The rule [`lib/rules/metadata/process.js`](https://github.com/w3c/specberus/blob/master/lib/rules/metadata/process.js) carries forward the value of `$('a#w3c_process_revision').attr('href')`, but that might be missing, or invalid.

(That happened recently: see [Echidna output](https://lists.w3.org/Archives/Public/public-tr-notifications/2016Sep/0061.html); see the [spec](https://rawgit.com/w3c/wcag/forchecking/understanding.html). The problem is that the spec uses `<a name="w3c_process_revision">` instead of `<a id="w3c_process_revision">`. Because that value is checked when trying to publish, and not before in the pipeline, Specberus did not complain.)

Perform some basic check in the rule.

tripu:
* Metadata extraction can now throw errors and warnings, like validation does, and those are captured by the application.
* Those events are shown on the web UI too, including rule desriptions and &ldquo;see rule in context&rdquo; and &ldquo;report a bug&rdquo; links (if pertinent).
* Throw an error if the process document isn't found (this fixes #501).
* One test document tweaked to fix this new error in it.

**Merge #504 before this one** (that PR is against this branch, not against `master`).

### DIFF
--- a/app.js
+++ b/app.js
@@ -43,6 +43,9 @@ app.use(express.static("public"));
 api.setUp(app);
 views.setUp(app);
 
+// @TODO Localise this properly when messages are translated; hard-coded British English for now.
+l10n.setLanguage('en_GB');
+
 server.listen(process.argv[2] || process.env.PORT || DEFAULT_PORT);
 
 io.sockets.on("connection", function (socket) {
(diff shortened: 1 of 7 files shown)


deniak:
I'm afraid the errors from the validation and the metadata extraction are getting mixed and the users can get confused, especially if these metadata are shown on the UI.
I do think the errors such as "data not found" should be displayed during the validation.

PR #504, opened 2016-09-16
When auto-detecting, show also extracted metadata in result table

### LINKED ISSUE #301: Show on results the extracted metadata
Once #299 is merged, the web UI of Specberus will need to invoke `extractMetadata()` first, and display the information returned as `info()` messages on the results table.

tripu:
Looks like this:

exp

(If it's too intrusive, we can limit its height, like this:

coll

Although I prefer not to: scrollbars within scrollbars are annoying and not a good practice.)

Fixes #301.

**Merge before #503** (this goes on top of those changes).

### DIFF
--- a/public/js/specberus.js
+++ b/public/js/specberus.js
@@ -196,11 +196,6 @@ jQuery.extend({
             if (data.warnings && data.warnings.length > 0)
                 for (var w in data.warnings)
                     addMessage(MSG_WARN, w);
-            const json = JSON.stringify(data.metadata, null, 4)
-            ,   jsonMessage = `<p>Review the <strong>metadata that was inferred from the document</strong>
-                to make sure that there are no errors:</p>\n<pre>${json}</pre>`
-            ;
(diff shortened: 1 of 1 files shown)


deniak:
Can't we just link to `/api/metadata`? This is not something useful most people using pubrules.
I think it's better to display only the important results on the UI. I'm pretty sure people only care about the errors and ignore the "advice".

tripu:
Too much info that will be ignored? Mmmh, maybe.

What if we add just a line next to (or below) the one that says *&ldquo;n rules were checked. Hover over the rows below for options&rdquo;*, saying something like *&ldquo;see the metadata that was extracted form this document&rdquo;* with a link to `api/metadata` for the document?

deniak:
:+1:

tripu:
OK, I'll modify the PR.
(No need to hold #503 now).

tripu:
Ready again, @deniak.
Hope this is what we both had in mind!

deniak:
@tripu Perfect!

Issue #493, opened 2016-09-05
Bug in rule “headers.copyright”: Permissive document license


jean-gui:
If a document uses the permissive document license, a warning is thrown (which might be normal) and the sample code is wrong.

Found [while checking `https://www.w3.org/TR/2016/CR-activitystreams-core-20160906/`](https://www.w3.org/pubrules/?url=https%3A%2F%2Fwww.w3.org%2FTR%2F2016%2FCR-activitystreams-core-20160906%2F&profile=CR&validation=simple-validation&noRecTrack=false&informativeOnly=false&echidnaReady=false&patentPolicy=pp2004).

tripu:
https://www.w3.org/2015/Process-20150901/ does not mention the permissive licence.

@deniak, shall we update https://www.w3.org/pubrules/doc/rules?profile=CR#copyright ? &ldquo;The copyright must use the following markup [&hellip;]&rdquo;. It's no *must*. And we should mention the &ldquo;permissive&rdquo; variant. Yes?

PR #551, opened 2016-11-29
Remove warning when using permissive licence

plehegar:
Probably fixes https://github.com/w3c/specberus/issues/493 at the same time...


plehegar:
@wseltzer is ok with removing the warning.

PR #587, opened 2017-03-06
Notes are *not* on REC-track

deniak:


deniak:
@tripu you are right! Notes are always informative. PR updated!

PR #591, opened 2017-03-07
Refactor texts of rules

tripu:
I extracted common factor from `lib/rules.json`. Not complete, but I think I identified most redundant rules. `rules.json` takes now 30% of the space; from 285 KB down to 85 KB (even if it has more lines now).

Some rules are parameterised, using placeholders `@{param1}`, `@{param2}`, etc.

Now, the content of  particular *{profile, section, rule}* can be one of:

* `"true"`: means &ldquo;go and take the text from special profile `*`&ldquo; (at the end of the file).
* A `string`: use that text.
* An `array`: &ldquo;go and take the text from special profile `*`, and interpolate placeholders with the strings provided&ldquo;.

Profiles are now &ldquo;ordered&rdquo;, from notes, to WDs, to RECs, to obsoleted/rescinded RECs. Also, better displayed now on `/doc/`. Please, review this order, too.

Old `rules-wrapper.js` is not needed any more; removed.

I'm *not* squashing commits because I want to preserve intermediate statuses in case I'm losing something in the process, and to keep my refactoring script at hand.


tripu:
@deniak, I can't remember now if this is exactly what you had in mind in #422&hellip;?

deniak:
@tripu after a quick look, this looks like a very nice improvement. I'm gonna need more time to review each profile though.
To answer your question, #422 is a different issue. Let's say you select the profile WD. That WD can either be on REC-track or not, that's why we have that checkbox on the interface (same thing for 'informative only'). Depending on that selection, the rule changes a little bit.

So if we wanted to do things correctly, there should be the same options in the rule description pages and the rules may change depending on the different options selected. Note, these options are not compatible with all profiles, eg. you can have a REC not on REC track.

tripu:
@deniak,

> *After a quick look, this looks like a very nice improvement. I'm gonna need more time to review each profile though.*

Thanks!

In case you find it useful, I reviewed them by opening the rules for the same profile in `w3.org/pubrules` and in my local copy (this branch) in different tabs, switching among them, and looking for differences visually. And same thing checking a few documents and examining the output (errors, etc) in both instances.

> *To answer your question, #422 is a different issue. Let's say you select the profile WD. That WD can either be on REC-track or not, that's why we have that checkbox on the interface (same thing for 'informative only'). Depending on that selection, the rule changes a little bit.*
> *So if we wanted to do things correctly, there should be the same options in the rule description pages and the rules may change depending on the different options selected. Note, these options are not compatible with all profiles, eg. you can have a REC not on REC track.*

OK, understood. I'll see if I can make that change, too.

deniak:
Checklist:
- [ ] FPIG-NOTE First Public Interest Group Note
- [ ]     IG-NOTE Interest Group Note
- [ ]     FPWG-NOTE First Public Working Group Note
- [ ]     WG-NOTE Working Group Note
- [ ]     FPWD First Public Working Draft
- [ ]     WD Ordinary Working Draft
- [ ]     FPCR First Public Candidate Recommendation
- [ ]     CR Candidate Recommendation
- [ ]     PR Proposed Recommendation
- [ ]     PER Proposed Edited Recommendation
- [ ]     REC Recommendation
- [ ]     RSCND Rescinded Recommendation
- [ ]     OBSL Obsolete Recommendation

lib/rules.json
```diff
@@ -1,1196 +1,1440 @@
...
-                        "knownDisclosureNumber": "It <span class=
```
deniak: s/NOTE/CR/

deniak: s/WD/CR/

deniak: s/CR/WD/

deniak: s/PR/NOTE/

deniak: s/NOTE/CR/

deniak: s/WD/CR/

deniak: s/CR/WD/

deniak: s/PR/NOTE/

Issue #567, opened 2017-01-09
Not all profiles have rule messages associated to them


tripu:
[`public/data/profiles.json`](https://github.com/w3c/specberus/blob/master/public/data/profiles.json) contains:

* `WD`
* `FPWD`
* `FPCR`
* `CR`
* `PR`
* `PER`
* `REC`
* `RSCND`
* `FPWG-NOTE`
* `WG-NOTE`
* `FPIG-NOTE`
* `IG-NOTE`
* `CG-NOTE`
* `MEM-SUBM`
* `TEAM-SUBM`

while [`lib/rules.json`](https://github.com/w3c/specberus/blob/master/lib/rules.json) has messages for:

* `WD`
* `FPIG-NOTE`
* `FPWD`
* `CR`
* `FPCR`
* `PR`
* `PER`
* `REC`
* `WG-NOTE`
* `FPWG-NOTE`
* `IG-NOTE`

ie, these 4 are missing all their messages: `RSCND`, `CG-NOTE`, `MEM-SUBM`, `TEAM-SUBM`.

This is causing errors like `l10n.message(): unknown profile code “MEM-SUBM”`.

tripu:
(Apart from solving this particular error, I should also check consistency between these two files automatically in [L10n tests](https://github.com/w3c/specberus/blob/master/test/l10n.js#L177-L211).)

tripu:
**Update:** `RSCND` will be fine after #578 is merged (and `OBSL` will be added in both places okay).

tripu:
@deniak, will you fill in sections and rules for `TEAM-SUBM`, `MEM-SUBM` and `CG-NOTE` now, or tell me how to do it?

deniak:
@tripu yes, I'll add the missing rules for these profiles.

deniak:
It was fixed a long time ago

PR #594, opened 2017-03-09
Improve management and display of tracks and profiles

tripu:
Mainly:

* Remove `/public/data/profiles.json` (was redundant, demanded AJAX on client).
* Avoid hard-coding list of profiles in `/lib/util.js`.
* Extend `/lib/rules.json` with more info.
* Improve the way tracks and profiles are displayed.
* Prepare the code to add sections &amp; rules of 3 missing profiles.

More details:

* Build list of profiles in `/lib/util.js` programmatically by reading the filesystem (case of certain files under `/lib/profiles/` changed to facilitate this: `UPPERCASE.js` for actual profiles; `lowercase.js` for ancestors and code).
* `/lib/rules.json` improved:
  * Added `TEAM-SUBM`, `MEM-SUBM`, `CG-NOTE` (**empty; no sections or rules yet**).
  * Reordered tracks and profiles.
* `/lib/profiles/`: fixed values of `exports.name` & `status` of `TEAM-SUBM` and `MEM-SUBM`.
* URL `/doc/`: displaying all profiles now (3 new ones), and grouped by tracks.
* URL `/`:
  * Injecting tracks &amp; profiles on the view now (ie, no AJAX; faster loading).
  * Profile drop-down: removed disabled pseudo-profile &ldquo;select an option&rdquo;; set auto by default (also, now all profiles fit in drop-down without a vertical scroll bar on my Chromium).
* Adapted code to all these changes: `app.js`, HTML, CSS, couple of templates, views and some other logic.
* [The extension is not necessary when `require`ing JSON files](https://nodejs.org/dist/latest-v6.x/docs/api/modules.html#modules_file_modules).

@deniak, I think all this prepares the code well to complete #567.

I don't know exactly what sections and rules those 3 profiles need. Let's do that on a separate PR?


lib/exceptions.js
```diff
@@ -1,4 +1,4 @@
-var records = require('./exceptions.json');
+var records = require('./exceptions');
```
deniak: isn't it loading `exceptions.js` instead of `exceptions.json`?

lib/profiles/CG-NOTE.js
```diff
@@ -11,7 +11,7 @@ exports.config = {
 
 var base = require("./base");
 exports.rules = base.insertAfter(
-                        require("./TR").rules
+                        require("./tr").rules
```
deniak: Why lowercasing TR?

lib/profiles/MEM-SUBM.js
```diff
@@ -1,7 +1,7 @@
 
 exports.name = "MEM-SUBM";
 exports.config = {
-    status:             "SUBM"
+    status:             "MEM-SUBM"
```
deniak: The short status for member submission is `SUBM` (see recent submissions at https://www.w3.org/Submission/)

lib/util.js
```diff
@@ -36,83 +26,106 @@ const buildJSONresult = function(err, warn, inf, metadata) {
...
-exports.profiles = profiles;
-exports.buildJSONresult = buildJSONresult;
-exports.processParams = processParams;
+fs.readdir('./lib/profiles/', (err, files) => {
+    if (err)
+        throw new Error(`Cannot build profiles in dir "/lib/profiles/": "${err}"`);
+    else {
+        const profiles = {};
+        files.forEach((i) => {
+            const valid = /(^[A-Z][A-Z\-]*[A-Z](\-Echidna)?)\.js$/.exec(i);
```
deniak: Ok, now I understand why you lowercased some profiles. I'm a bit skeptical on adding that kind of requirement though.
What do you think about moving the "valid" profiles to `lib/profiles/TR/` and `lib/profiles/SUBM/` and leave the base profiles like `base.js`, `TR.js`, etc in `lib/profiles/`?

lib/exceptions.js
tripu: You're right. It should load that one. I think it is *not* loading that one in this case because the loader is `exceptions.js` itself, so it detects the infinite loop and loads the `.json` instead (this particular casuistic [is hinted](https://nodejs.org/dist/latest-v6.x/docs/api/modules.html#modules_cycles), but not properly documented&nbsp;&mdash;&nbsp;or I can't find it).

lib/profiles/MEM-SUBM.js
tripu: OK!
I see both [member](https://www.w3.org/Submission/) and [team](https://www.w3.org/TeamSubmission/) submissions use `SUBM`. Shall I correct the `status` of `lib/profiles/TEAM-SUBM.js`, too (I set it to `TEAM-SUBM`)? (Is it OK that both profiles share the same *short status*?)

lib/util.js
tripu: You're right: it's an ugly hack. I'll restore uppercases and move profiles under those two subdirs you suggest.

lib/util.js
```diff
@@ -36,83 +26,111 @@ const buildJSONresult = function(err, warn, inf, metadata) {
...
-        else if ('profile' === p) {
-            // Profile: if it's a string, load the corresponding object.
-            if (result.hasOwnProperty(p))
-                throw new Error('Parameter “' + p + '” is used more than once');
-            else if ('string' === typeof params[p]) {
-                if (profiles.hasOwnProperty(params[p]))
-                    result[p] = profiles[params[p]];
-                else if ('auto' === params[p])
-                    result[p] = 'auto';
+const processParams = function(profiles) {
```
deniak: Can you give that `const` a different name so we don't confuse `processParams` and `exports.processParams`?

tripu:
I changed the name of that const (function), @deniak.

deniak:
:+1:

PR #604, opened 2017-03-20
headers: relax previous versions detection

deniak:


tripu:
[Tested automatically](https://coveralls.io/builds/10677560/source?filename=lib%2Fvalidator.js#L286)

PR #613, opened 2017-04-20
Fix relative links to profile rules

tripu:
Relative links on the [pubrules documentation page](https://www.w3.org/pubrules/doc/) to the different profiles are broken.

I think this fixes the issue.

### DIFF
--- a/views/doc/index.handlebars
+++ b/views/doc/index.handlebars
@@ -13,7 +13,7 @@
       <h4 id="{{this.abbr}}"><a href="#{{this.abbr}}">{{this.name}}</a></h3>
       <ul>
       {{#each this.profiles}}
-        <li><a href="{{../BASE_URI}}doc/rules/?profile={{this.abbr}}"><span class="profile-abbr">{{this.abbr}}</span> {{this.name}}</a></li>
+        <li><a href="{{BASE_URI}}doc/rules/?profile={{this.abbr}}"><span class="profile-abbr">{{this.abbr}}</span> {{this.name}}</a></li>
       {{/each}}
       </ul>
     {{/each}}
--- a/views/doc/index.handlebars
+++ b/views/doc/index.handlebars
@@ -13,7 +13,7 @@
       <h4 id="{{this.abbr}}"><a href="#{{this.abbr}}">{{this.name}}</a></h3>
       <ul>
       {{#each this.profiles}}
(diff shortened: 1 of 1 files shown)


tripu:
Oops. Or maybe not.
Let me double-check, @deniak&hellip;

tripu:
Ready, @deniak. I think this is the right way.

(For [our production instance](https://www.w3.org/pubrules/), we have to launch Specberus with `BASE_URI='/pubrules'`; I think we do already, but I just added that to our internal documentation.)

tripu:
@deniak, the issue is that `{{../BASE_URI}}` cannot be resolved by Handlebars in this context, because it refers to the variable `BASE_URI` that is being passed to the template, but [here it is used nested two levels deep](https://github.com/w3c/specberus/blob/master/views/doc/index.handlebars#L16) (because of those two `{{#each}}` blocks). See [Handlebar paths](http://handlebarsjs.com/#paths).

`{{../../BASE_URI}}` is what we need here (the parent of the parent context).

`../` is not used here to build a relative link; it's Handlebars syntax.

That the resulting relative link is okay when we omit the trailing slash from the URI is just a coincidence in this case.

deniak:
@tripu ok got it. Strange behavior though

tripu:
Thanks, @deniak.

---

For the record, this is what's happening.

* Because in this context `{{../BASE_URI}}` isn't valid, it resolves to `''` (empty string).
* `<a href="{{../BASE_URI}}doc/rules/?profile=FOO">` thus produces this hyperlink: `<a href="doc/rules/?profile=FOO">` (a [path-relative URL](https://url.spec.whatwg.org/#path-relative-url-string)).
* If you visit `https://www.w3.org/pubrules/doc/` (with slash), the relative URI is resolved as `https://www.w3.org/pubrules/doc/doc/rules/?profile=FOO`, which is wrong.
* On the other hand, if you visit `https://www.w3.org/pubrules/doc` (no slash), the relative URI is resolved as `https://www.w3.org/pubrules/doc/rules/?profile=FOO` (because the algorithm [&ldquo;remove[s] url’s path’s last item, if any&rdquo;](https://url.spec.whatwg.org/#relative-state)), which works.

This issue isn't specific to Specberus, but happens on any page that works equally with and without a trailing slash, and has hyperlinks of the form path-relative-URL.

PR #631, opened 2017-05-19
Require a mailing list (+ archive) or a github issue list

### LINKED ISSUE #629: Don't require mailing list
When [trying to publish](https://lists.w3.org/Archives/Public/public-tr-notifications/2017May/0075.html), I'm hitting:

```JS
      "specberus": {
        "status": "failure",
        "errors": [
          {
            "key": "no-arch",
            "type": {
              "name": "sotd.mailing-list"
            }
          }
        ]
      },
```

However, we are not using a mailing list for our spec feedback (nor do we want to use it for CR). This requirement should be relaxed.

deniak:
Fix #629


lib/rules/sotd/mailing-list.js
```diff
@@ -45,8 +45,8 @@ exports.check = function (sr, done) {
         if (!foundRepo) {
```
plehegar: if this is a REC, we ought to require a Repo rather than allowing the use of mailing list. This will align with the maintenance discussion within the AB.

deniak: That means changing the requirements for RECs. Can we do that without consulting spec-prod before?

plehegar: we should let spec-prod know but this is a necessary change however.

plehegar: Btw, this will be part of the transition requirements update at https://w3c.github.io/transitions/?docstatus=recommendation

PR #632, opened 2017-05-22
update the copyright regex to avoid dummy copyrights

deniak:
The current regex won't return an error on copyrights like:
```
Copyright © 2017 FOOBAR & W3C® (MIT, ERCIM, Keio, Beihang). W3C liability, trademark and permissive document license rules apply. 
```

These cases should be handled as exceptions.

### DIFF
--- a/lib/rules/headers/copyright.js
+++ b/lib/rules/headers/copyright.js
@@ -10,7 +10,7 @@ exports.check = function (sr, done) {
     if ($c.length) {
         var year = (sr.getDocumentDate() || new Date()).getFullYear()
         ,   text = sr.norm($c.text())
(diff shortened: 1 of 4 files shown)


lib/rules/headers/copyright.js
```diff
@@ -10,7 +10,7 @@ exports.check = function (sr, done) {
     if ($c.length) {
         var year = (sr.getDocumentDate() || new Date()).getFullYear()
         ,   text = sr.norm($c.text())
-        ,   startRex = "^Copyright (©|&copy;) (?:(?:199\\d|20[01]\\d)-)?" + year + " .*?W3C(®|&reg;) \\(MIT, ERCIM, Keio, Beihang\\)"
+        ,   startRex = "^Copyright (©|&copy;) (?:(?:199\\d|20[01]\\d)-)?" + year + " W3C(®|&reg;) \\(MIT, ERCIM, Keio, Beihang\\)"
```
plehegar: should we allow more whitespaces before W3C?

deniak: done

PR #634, opened 2017-05-24
RECs now require a github repo issue link instead of a mailing list + archive

deniak:


tripu:
(Note to self: inform spec-prod ML when it's deployed in production, as with #629.)

PR #676, opened 2017-09-19
Link rel=canonical

deniak:
Starting from Oct 1, we will require a `rel=canonical`.
See https://lists.w3.org/Archives/Public/spec-prod/2017JulSep/0005.html for more details.

### DIFF
--- a/lib/rules/structure/canonical.js
+++ b/lib/rules/structure/canonical.js
@@ -15,7 +15,7 @@ exports.check = function (sr, done) {
     // See https://lists.w3.org/Archives/Public/spec-prod/2017JulSep/0005.html
     sr.transition({
         to:          new Date('2017-09-30')
-    ,   doMeanwhile: function () { return done(); }
+    ,   doMeanwhile: () => {}
     ,   doAfter:     checkCanonical
     });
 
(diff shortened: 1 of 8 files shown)


lib/rules/structure/canonical.js
```diff
@@ -0,0 +1,23 @@
...
+    const checkCanonical = function () {
+        var $lnk = sr.$("head > link[rel=canonical]");
+        if (!$lnk.length || !$lnk.attr("href")) sr.error(self, "not-found");
+    };
+
+    // That canonical link is mandatory starting from Oct 1, 2017.
+    // See https://lists.w3.org/Archives/Public/spec-prod/2017JulSep/0005.html
+    sr.transition({
+        to:          new Date('2017-09-30')
+    ,   doMeanwhile: function () { return done(); }
```
tripu: Better

```js
    ,   doMeanwhile: () => {}
```

, as `done()` is called anyway before the end of function `check`.

Not sure if calling `done()` twice will distort the counter of &ldquo;pending checks&rdquo; and thus the progress bar on the UI. But just in case, and for simplicity.

deniak: I updated the code with your suggestion! Thanks.

tripu:
Merging, as the build errors are not specific to this branch (nsp error, affecting all branches now).

PR #711, opened 2017-12-01
Skip CSS validation for the CSS specs

deniak:
It was suggested that we skip the CSS validation or at least ignore the errors for the CSS specs.
That PR adds that exception for all the specs with a shortname starting with `css`.

/cc @svgeesus


deniak:
It's live now!

PR #716, opened 2017-12-15
Handle submissions

### LINKED ISSUE #366: Member submissions style
TR now benefits from a new style. Should Member submissions migrate as well?
@plehegar WDYT?

deniak:
That PR adds:
* the documentation for the Member and Team Submissions
* align rules according to their descriptions

Fix #366 #397 

cc @plehegar @wseltzer


deniak:
> When testing against the most recent [team submission](https://www.w3.org/TeamSubmission/) and most recent [member submission](https://www.w3.org/Submission/), in both cases I got, among the rest of messages, a *warning* (in orange) with the only text &ldquo;undefined&rdquo;. Looked like some `sr.warning()` receiving the wrong metadata for some reason. I tried to debug it, but could not reproduce it. Weirdly, it seems to happen only the first time you submit a document to Specberus once it's started. Subsequent checks don't show that message (?). Can you reproduce that by chance, @deniak?

Yes, good catch. That bug only occurs the first time because we are trying to extract the metadata from the document in the 'auto' mode. I fixed the issued by removing an unnecessary warning in `metadata.process`.

> Running Specberus against those two documents (most recent team submission and most recent member submission) produces quite a few errors and warnings in both cases. Do all those messages look legit to you, @deniak? (Looks like those two would not pass pubrules, if they were to be published today&hellip;)

Indeed, given the low number of submissions, they are easily outdated with the recent changes in pubrules but I looked at the errors and they are good errrors to me (switch URLs to https, etc).

tripu:
Thanks for that change, @deniak.
Merge whenever you're ready; we can tag and deploy right afterwards.

PR #752, opened 2018-05-18
Update to node 10 in .travis.yml

tripu:


package.json
```diff
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "compression": "1.7.2",
-    "express": "5.0.0-alpha.6",
+    "express": "4.16.3",
```
deniak: Why using an older version?

tripu: Mmmh&hellip; seems safer to stay away from alphas and prereleases until they're well tested?
That being said, I realise it was me who accepted that Greenkeeper update! :laughing:
Whatever you prefer.

deniak: I prefer to stick with stable versions also

PR #754, opened 2018-05-24
fix(package): update request to version 2.87.0

tripu:
Closes #748


tripu:
Tested automatically in [`lib/rules/headers/dl.js#L52`](https://coveralls.io/builds/17143523/source?filename=lib%2Frules%2Fheaders%2Fdl.js#L52).

PR #570, opened 2017-01-17
[On hold] Drop rule about link to translations

tripu:
Closes #552, closes #730.

PR #757, opened 2018-06-04
fix translation recommended text

deniak:
The recommended text for the translation warning doesn't match the rule.


tripu:
#570 made this one unnecessary

PR #763, opened 2018-07-12
min review date is 28 days + handle editorial updates

deniak:
The minimum review period for CR and PR is 28 days per https://www.w3.org/2018/Process-20180201/.
That PR retrieves only dates between the publication date +28 days and +1 year.

Also, for CR updates with editorial changes, we don't need to check that date so we simply skip the rule.


deniak:
@tripu it's not specified in the process but it's very unlikely the review can last more than 1 year. AFAIK, it never happened.

PR #764, opened 2018-07-13
add 'editorial' in the list of allowed paramaeters

deniak:


deniak:
> * Does it make sense to extend the options in the GUI, to have &ldquo;editorial changes&rdquo; along with &ldquo;not on rec-track&rdquo;, &ldquo;informative-only&rdquo;, etc?
> * Shall we document this new param [on the README](https://github.com/w3c/specberus#4-js-api), too?

That parameter is actually a bit tricky as it applies only to CR updates. I think there should be another profile instead. I'll merge and create another profile in a separate PR.

PR #776, opened 2018-10-08
Suggest a <script> block that works for XHTML5, too

tripu:
[Rule #fixupJs](https://www.w3.org/pubrules/doc/rules/?profile=REC#fixupJs)

cf https://github.com/w3c/specberus/pull/297#issuecomment-427820666

/cc @Mouvedia

### DIFF
--- a/lib/rules.json
+++ b/lib/rules.json
@@ -1934,7 +1934,7 @@
                     "compoundFilesLocation": "If the document is compound (i.e., if it consists of more than one file), all the files <span class=\"rfc2119\">must</span> be under a directory /TR/YYYY/@{param1}-shortname-YYYYMMDD/",
                     "compoundOverview": "The main page <span class=\"rfc2119\">should</span> be called Overview.html.",
                     "compound": "All other files <span class=\"rfc2119\">must</span> be reachable by links from the document.",
-                    "fixupJs": "The document <span class=\"rfc2119\">must</span> include the script <code><a href=\"https://www.w3.org/scripts/TR/2016/fixup.js\">fixup.js</a></code>.<p><span style=\"font-style: italic\">Include this source code:</span><br/><code>&lt;script src=\"//www.w3.org/scripts/TR/2016/fixup.js\" type=\"text/javascript\"&gt;&lt;/script&gt;</code></p>"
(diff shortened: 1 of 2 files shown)


Mouvedia:
The most important is to go and fix the specifications that are served as XML.

lib/rules.json
```diff
@@ -1934,7 +1934,7 @@
...
                     "compoundOverview": "The main page <span class=\"rfc2119\">should</span> be called Overview.html.",
                     "compound": "All other files <span class=\"rfc2119\">must</span> be reachable by links from the document.",
-                    "fixupJs": "The document <span class=\"rfc2119\">must</span> include the script <code> <a href=\"https://www.w3.org/scripts/TR/2016/fixup.js\">fixup.js</a></code>.<p><span style=\"font-style: italic\">Include this source code:</span><br/><code>&lt;script src=\"//www.w3.org/scripts/TR/2016/fixup.js\"&gt;&lt;/script&gt;</code></p>"
+                    "fixupJs": "The document <span class=\"rfc2119\">must</span> include the script <code><a href=\"https://www.w3.org/scripts/TR/2016/fixup.js\">fixup.js</a></code>.<p><span style=\"font-style: italic\">Include this source code:</span><br/><code>&lt;script src=\"//www.w3.org/scripts/TR/2016/fixup.js\" type=\"text/javascript\"&gt;&lt;/script&gt;</code></p>"
```
deniak: According to the [rfc4329](https://tools.ietf.org/html/rfc4329)`text/javascript` is obsolete now. It should be `application/javascript`

Mouvedia: …which is incompatible with Internet Explorer ≤8.

tripu: Thanks, @deniak &amp; @Mouvedia; you both taught me something I didn't know.

Anyway, I was already &ldquo;fixing&rdquo; this (see last commit).

[According to caniuse.com](https://caniuse.com/usage-table), all versions of IE &le;8 account for &lt;0.2% of market share. I think IE 8 is now past the point where we should accommodate it or sacrifice features for it&hellip;

That being said, I leave it in the hands of the reviewer to decide whether to take this fix, or revert my last commit :innocent:

PR #781, opened 2018-11-07
wpub exception

deniak:
That PR adds an exception for wpub
Closes https://github.com/w3c/echidna/issues/596

### DIFF
--- a/lib/exceptions.js
+++ b/lib/exceptions.js
@@ -33,11 +33,10 @@ Exceptions.prototype.has = function (shortname, rule, key, extra) {
     for (var i = set.length - 1; i >= 0; i--) {
         for (var y = set[i].length - 1; y >= 0; y--) {
             var exception = set[i][y];
-
             if (rule === exception.rule &&
                 (extra === undefined ||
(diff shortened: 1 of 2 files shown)


lib/exceptions.js
```diff
@@ -33,11 +33,10 @@ Exceptions.prototype.has = function (shortname, rule, key, extra) {
     for (var i = set.length - 1; i >= 0; i--) {
         for (var y = set[i].length - 1; y >= 0; y--) {
             var exception = set[i][y];
-
             if (rule === exception.rule &&
                 (extra === undefined ||
                   (!exception.message && compareValue(extra.type, exception.type)) ||
-                  (exception.message && compareValue(extra.type, exception.type) && compareValue(extra.message, exception.message)))) {
+                  (exception.message && compareValue(extra.message, exception.message)))) {
                 return true;
```
tripu: Why this change?

deniak: The CSS validator differentiates the different types of errors, but not the markup validator so I think looking for the exceptions by type *or* message is enough and the system should be generic enough to work with both validators.

PR #788, opened 2018-11-26
Update all dependencies before tagging, releasing and publishing a new version

tripu:


tripu:
I see a new warning on the console (*&ldquo;node:4493 ExperimentalWarning: The http2 module is an experimental API&rdquo;*)&nbsp;&mdash;&nbsp;I suspect coming from either `express`, `eslint` or `eslint-plugin-node`. But apart from that, tests pass, and a very quick test of the web UI and the API shows no problems.

PR #798, opened 2019-01-22
'npm run audit': make more robust (improve regex)

tripu:
The issue was a recent switch from vulns of the form
`https://nodesecurity.io/advisories/xxx`
to
`https://npmjs.com/advisories/xxx`
:P

This should fix it.

Closes #796


deniak:
Nice catch @tripu

PR #851, opened 2019-11-26
webauthn-2 use markup in its <h1> element

plehegar:
[[
Content of title and h1 do not match.
The document's title must be in the title element and in an h1 element.
]]
from [pubrules](https://www.w3.org/pubrules/?url=https%3A%2F%2Fwww.w3.org%2FTR%2F2019%2FWD-webauthn-2-20191126%2F&profile=WD&validation=simple-validation&noRecTrack=false&informativeOnly=false&echidnaReady=false&patentPolicy=pp2004)

cc @akshayku @wseltzer

### DIFF
--- a/lib/exceptions.json
+++ b/lib/exceptions.json
@@ -36,7 +36,7 @@
   ],
   "^webauthn-2$": [
     {
-      "rule": "metadata.title"
+      "rule": "headers.h1-title"
     }
   ],
   "^did-core$": [
--- a/lib/exceptions.json
+++ b/lib/exceptions.json
@@ -34,6 +34,11 @@
       "rule": "headers.copyright"
     }
   ],
+  "^webauthn-2$": [
+    {
+      "rule": "metadata.title"
+    }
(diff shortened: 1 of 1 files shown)


deniak:
For that specific use case, I feel like it would be better to sanitize the title before checking it. I think it's ok to remove the line breaks while fetching the title. If you are ok with my suggestion, I'll submit a PR.

deniak:
Looking closely, removing the line break is not enough since we have the punctuation.

lib/exceptions.json
```diff
@@ -34,6 +34,11 @@
       "rule": "headers.copyright"
     }
   ],
+  "^webauthn-2$": [
+    {
+      "rule": "metadata.title"
```
deniak:
```suggestion
      "rule": "headers.h1-title"
```

plehegar:
Added the change. Thank you Denis.

deniak:
Merged and deployed

PR #852, opened 2019-11-28
Our First GitHub action to greet new contributors

vivienlacourba:


deniak:
We currently don't have any special requirements for first time contributors. We'll revisit after we come up with some coding styles and contributor guidelines.

PR #853, opened 2019-11-28
GitHub Action to mark stale issues and pull requests

vivienlacourba:


deniak:
That feature is currently not wanted as issues and staled PR would need to be checked manually before closing.

PR #863, opened 2020-01-29
misc fixes reported by CI

deniak:


deniak:
@dontcallmedom eslint started to report an error on the `x.hasOwnProperty(y)`. See https://eslint.org/docs/rules/no-prototype-builtins

dontcallmedom:
https://eslint.org/docs/rules/no-prototype-builtins#when-not-to-use-it says:
> You may want to turn this rule off if your code only touches objects with hardcoded keys, and you will never use an object that shadows an Object.prototype method or which does not inherit from Object.prototype.

I haven't review the full specberus architecture, but from a quick glance, it seems we're only dealing with plain-old regular JS objects here, with no risk of anything fancier coming in (hence why I'm not even sure we need the `hasOwnProperty` approach in the first place)

deniak:
Actually, we had some issues in the past because of the way we managed the documentation of each rule. There's a [json](https://github.com/w3c/specberus/blob/master/lib/rules.json) listing the documentation per profile and we missed some keys in the past. This explains the multiple calls to `hasOwnProperty`. While this can probably be improved, the syntax suggested by eslint seems safer.

dontcallmedom:
I'm still confused on how the `hasOwnProperty` syntax is helping, but I'm not standing in the way :)

Issue #766, opened 2018-07-17
Pubrules silently crashes when submitting a non TR document with the 'auto' profile


deniak:
For instance, checking https://www.w3.org/ won't produce any error. We should handle the exception and return a proper error.

deniak:
Fixed by https://github.com/w3c/specberus/pull/879

PR #869, opened 2020-03-04
Add check of non-TR document

nissy-insfc:
Bug fix for issue #766

### DIFF
--- a/app.js
+++ b/app.js
@@ -49,6 +49,8 @@ io.sockets.on("connection", function (socket) {
     socket.emit("handshake", { version: version });
     socket.on("extractMetadata", function (data) {
         if (!data.url) return socket.emit("exception", { message: "URL not provided." });
+        if (!(/^https?:\/\/www.w3.org\/TR\/.+/.test(data.url)))
+            return socket.emit("exception", { message: "Invalid URL of document." });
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -49,6 +49,8 @@ io.sockets.on("connection", function (socket) {
     socket.emit("handshake", { version: version });
     socket.on("extractMetadata", function (data) {
         if (!data.url) return socket.emit("exception", { message: "URL not provided." });
+        if (!(/^https?:\/\/www.w3.org\/TR\/.+/.test(data.url)))
+            return socket.emit("exception", { message: "Invalid URL of document." });
```
deniak: Although the check seems fine, it doesn't handle all the use cases:

1. Someone can still try to check the following document (https://www.w3.org/TR/2019/NOTE-wai-aria-practices-1.1-20190207/examples/listbox/listbox-rearrangeable.html) and the exception will break pubrules as it's not the main page of a specification.
2. People may want to check their document *before* uploading them to /TR. In this case, it would be staged somewhere else (like github pages) and pubrules will not run because of the check you added.

You need to check what exception pubrules is generating when it's checking a non-TR doc, catch it and return the error.

deniak:
Will not fix

PR #879, opened 2020-04-07
FIX#766 Add non-TR check, Add live instructions to README.md

jennyliang220:


lib/rules/metadata/dl.js
```diff
@@ -56,21 +56,25 @@ exports.check = function(sr, done) {
...
-          return done(result);
+            .set("User-Agent", ua);
+        req.query({ apikey: apikey });
+        req.end(function (err, res) {
+            result.updated = !(err || !res.ok);
+            return done(result);
         });
+    } else {
+        return done();
+    }
```
deniak, line 79:
```suggestion
    } else {
        throw new Error('[EXCEPTION] The document date could not be parsed.');
    }
```

lib/validator.js
```diff
@@ -218,6 +218,10 @@ Specberus.prototype.getDocumentDate = function () {
             return false; // no need to look further
         }
     });
+
+    if (!this.docDate) {
+        return this.throw("The URL should be a TR document.");
+    }
```
deniak, line 224: Thinking more about it, I think this should probably be handle when the rules are being checked (i.e. in `Specberus.prototype.extractMetadata` (see at the end of that comment)

```suggestion
```
```
        profileMetadata.rules.forEach(function (rule) {
            try {
              rule.check(self, function (result) {
                  if (result) {
                      for (var i in result) {
                          self.meta[i] = result[i];
                      }
                  }
                  done++;
                  self.sink.emit("done", this.name);
                  if (done === total)
                      self.sink.emit("end-all", util.buildJSONresult(errors, warnings, infos, self.meta));
              }.bind(rule));
            } catch (e) {
                self.throw(e.message);
            }
        });
```

package.json
```diff
@@ -43,6 +43,7 @@
     "coverage": "istanbul cover _mocha",
     "coveralls": "npm run coverage && cat ./coverage/lcov.info | coveralls",
     "build": "npm run lint && npm run hint && npm run test && npm run jsdoc",
+    "live": "nodemon --use_strict app",
```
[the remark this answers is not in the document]
jennyliang220: fixed in https://github.com/w3c/specberus/pull/879/commits/a4235b87713016fac72b29040b94a486ba9cb786

README.md
```diff
@@ -75,6 +77,15 @@ $ BASE_URI=https://spec-store.com/check/ W3C_API_KEY=deadbeef npm start
 $ BASE_URI=/hostname/can/be/omitted/ W3C_API_KEY=deadbeef npm start 88
 ```
 
+## 2. Auto reload when developing
+
+Use `npm install -g nodemon` to install `nodemon` globally, then run `npm run live`.
```
jennyliang220, line 82: no need to install

PR #882, opened 2020-04-14
FIX#849 FIX#827 Change Patent Policy checks

### LINKED ISSUE #849: Note and Patent Policy wording
specberus expects:

`<p data-deliverer="@@wgId@@">This document was produced by a group
operating under the <a href="https://www.w3.org/Consortium/Patent-Policy/"><abbr title="World Wide Web Consortium">W3C</abbr> Patent
Policy&lt;/a>.</p>`

but [respec](https://github.com/w3c/respec/blob/develop/src/w3c/templates/sotd.js#L222) adds in the same paragraph:

`The group does not expect this document to become a W3C Recommendation.`

not sure how to modify https://github.com/w3c/specberus/blob/master/lib/rules/sotd/pp.js#L25 but it would be nice to relax it.

jennyliang220:
closes #881 #849


lib/rules/sotd/pp-note.js
```diff
@@ -0,0 +1,39 @@
...
+            // a Working Group Note || an Interest Group Note
+        ,   keyword = isWGNote ? 'a Working' : 'an Interest'
+        ,   wantedText = "^Publication as " + keyword + " Group Note does not imply " +
+            "endorsement by the W3C Membership. This is a draft document " +
+            "and may be updated, replaced or obsoleted " +
+            "by other documents at any time. It is inappropriate to cite this document as other than work in progress.$"
+        ,   wantedReg = new RegExp(wantedText)
+        ;
+        console.log('wantedText: ');
+        console.log(wantedText);
```
deniak, line 23: remove debug log

lib/rules/sotd/pp.js
```diff
@@ -48,7 +43,7 @@ function buildWanted (groups, sr) {
                 wanted += " This Amended Recommendation was produced by incorporating errata after that group closed\\.";
         }
     }
-    wanted = '^' + wanted + '$';
+    wanted = wanted ? '^' + wanted + '$' : '';
```
deniak, line 46: That rule becomes useless for IG Notes, so I think we should remove it from the "IG Note" profile and `wanted` should never be empty.

lib/rules/sotd/pp-note.js
```diff
@@ -0,0 +1,39 @@
...
+            var $p = sr.$(this)
+            ,   text = sr.norm($p.text())
+            ;
+            if (wantedReg.test(text)) {
+                findPPNote = true;
+                return false;
+            }
+        });
+        if (!findPPNote) {
+            sr.info(self, "expected", { "text": wantedText });
```
deniak, line 35: With only the `sr.info`, pubrules will not return an error if the paragraph is missing. We should add the error and its message in `l10n-en_GB.js` like:

```suggestion
            sr.info(self, "expected", { "text": wantedText });
            sr.error(self, "no-notepp");
```

PR #886, opened 2020-04-24
Fixed #871. Change getFeedbackDueDate, split 'date found' and 'valid date'

### LINKED ISSUE #871: "Minimal CR end date not found" message should be more explicit
When checking a CR, some people complained that it was not able to find the CR end date while it was actually in the SotD.
The minimal review end date is at least 28 days starting from the publication date.
If the period found is less that 28 days, the error should say so and not simply "end date not found".

jennyliang220:
Change getFeedbackDueDate, split 'date found' and 'valid date'


test/rules.js
```diff
@@ -392,6 +392,7 @@ var tests = {
         ]
     ,   'cr-end':  [
             { doc: "metadata/cr-mediacapture-streams.html", config: { status: "CR" }}
+    ,       { doc: "online/CR-JSON-ld.html", config: { status: "CR"}, errors: ["sotd.cr-end"] } //sotd.cr-end.found-not-valid
```
deniak, line 395: It would probably be better to add smaller test documents like in `test/docs/sotd` and test other use cases:
* feedback due date detected < publication date
* feedback due date detected > publication date but only 27 days 
* feedback due date detected > publication date with more than 28 days

It's not useful to add a 13k lines documents to test only one paragraph.

PR #891, opened 2020-04-30
Fix #873, Check previous version for Notes unless they are First Public Note

### LINKED ISSUE #873: Check previous version for Notes unless they are First Public WG Note
At the moment, we don't expect Notes to have a previous version. However, this may happen and if it was previously published, we need to make sure the previous version is specified and is the right one.

jennyliang220:


lib/rules/metadata/profile.js
```diff
@@ -68,8 +72,46 @@ exports.check = function(sr, done) {
...
+            var $linkLatest = (dts.Latest) ? dts.Latest.dd.find("a").first().attr('href') : '';
+            shortname = $linkLatest.match(/^https:\/\/www.w3.org\/TR\/(.+)\/$/)[1];
+            var req = sua.get('https://api.w3.org/specifications/' + shortname)
+                .set("User-Agent", "W3C-Pubrules/" + sr.version);
+            req.query({ apikey: process.env.W3C_API_KEY });
+            req.end(function (err, res) {
+                if (err || !res.ok) {
+                    if (res && res.status == 404) {
+                        specExists = false;
+                    }
```
deniak, line 96: There could actually be other kind of errors other than 404 (e.g. network error, 500, etc). I guess in this case, we can simply assume the spec doesn't exist.

```suggestion
                    specExists = false;
```

lib/rules/metadata/profile.js
```diff
@@ -68,8 +72,46 @@ exports.check = function(sr, done) {
...
+        return new PowerPromise(function (resolve) {
+            var $dl = sr.$("body div.head dl").first()
+            ,   dts = sr.extractHeaders($dl)
+            ,   $linkPrev = (dts.Previous) ? dts.Previous.dd.find("a").first() : ''
+            ,   shortname
+            ,   specExists
+            ;
+
+            var $linkLatest = (dts.Latest) ? dts.Latest.dd.find("a").first().attr('href') : '';
+            shortname = $linkLatest.match(/^https:\/\/www.w3.org\/TR\/(.+)\/$/)[1];
```
deniak, line 88: Check if the index 1 actually exists. There might be a parsing issue

lib/rules/metadata/profile.js
```diff
@@ -68,8 +72,46 @@ exports.check = function(sr, done) {
...
+                } else {
+                    specExists = true;
+                }
+                resolve(specExists || $linkPrev);
+            });
+        });
+    };
+
+    if (id) {
+        if (/-NOTE/.test(id)) {
```
deniak, line 106:
```suggestion
        if (/\-NOTE/.test(id)) {
```

jennyliang220: After adding escape `\`, build failed: https://travis-ci.com/github/w3c/specberus/builds/166576468

```
/home/travis/build/w3c/specberus/lib/rules/metadata/profile.js
  105:14  error  Unnecessary escape character: \-  no-useless-escape
```

deniak: Damn `eslint`! Ok let's skip that "useless" escape :)

Issue #706, opened 2017-11-11
Add an option to return the actual error values


marcoscaceres:
Right now, the returned errors, etc. are just ids (e.g., "links.linkchecker"). It would be nice to either return the actual error description, or have an API to look up a batch of error ids to get back the english text.

tripu:
(In the meantime, the &ldquo;manual lookup API&rdquo; goes more or less like this: ➀ open [`lib/l10n-en_GB.js`](https://github.com/w3c/specberus/blob/master/lib/l10n-en_GB.js); ➁ press `⎈`+`F`; ➂ enter the error ID, eg `links.linkchecker` and find the human-readable message.)

binji:
It would be nice to have the additional information for validators too; this is what I get currently:
```
...
{"name":"validation.css","section":"document-body","rule":"cssValide"},
{"name":"validation.html","section":"format","rule":"valideHTML"},
{"name":"validation.html","section":"format","rule":"valideHTML"},
{"name":"validation.html","section":"format","rule":"valideHTML"},
{"name":"validation.html","section":"format","rule":"valideHTML"},
{"name":"validation.html","section":"format","rule":"valideHTML"},
{"name":"validation.html","section":"format","rule":"valideHTML"},
...
```

I ended up hacking `rule/validation/html.js`. to log the full errors, e.g.

```
{ line: 4983,
  column: 2839,
  message: 'Duplicate attribute “style”.',
  link: 'https://validator.w3.org/nu/?doc=null' }
```

sidvishnoi:
To get proper message from API results, we need access to `key`.

Looking at the code, I think it's a bug that `key` and `extra` info are not available.
https://github.com/w3c/specberus/blob/6321df0ce152d7d56357d39f7f8253a256a91b0b/lib/validator.js#L149
https://github.com/w3c/specberus/blob/6321df0ce152d7d56357d39f7f8253a256a91b0b/lib/validator.js#L106

It should probably be something like:
```
 self.sink.on('err', function (...data) { errors.push(data); }); 
```
...but probably flattened into a single object.

Was trying to integrate this with ReSpec today.

PR #893, opened 2020-05-05
flatten event params for err, warning and info

deniak:
suggedted by @sidvishnoi in https://github.com/w3c/specberus/issues/706.

We actually rely on the 2 `emit` params for the documentation so I ended up merging both of them in the listeners.
Does that solve your issue @sidvishnoi ?

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -70,8 +70,8 @@ const processRequest = function(req, res) {
         if (validate && 'auto' === options.profile) {
             errors = [];
             v = new validator.Specberus();
-            handler = new Sink(function(data) {
-                errors.push(data);
+            handler = new Sink(function(...data) {
+                errors.push(Object.assign({}, ...data));
             }, function(data) {
                 if (errors.length > 0)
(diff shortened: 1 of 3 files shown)


sidvishnoi:
Thanks @deniak! This does solve the problem of not having `key` and `extra` in API results. Still need to manually convert messages using [lib/l10n.js](https://github.com/w3c/specberus/blob/master/lib/l10n.js) though. Would be nice if API returned full message, but this is good for now.

But I'm not sure if mutating the input (modifying `data[0]` of `...data`) in a listener is a good idea. I don't know why `Object.assign({}, ...data)` didn't work.

deniak:
@sidvishnoi after I read your comment, I started investigating on why the `assign` on an empty object didn't have the expected result and there's indeed a side effect if we reassign the `data`.
I actually missed the listener in the API results which rely on [`Sink`](https://github.com/w3c/specberus/blob/master/lib/sink.js) so this should now be better.

PR #942, opened 2020-08-04
Add https check for links

### LINKED ISSUE #906: Editor's draft, implementation report, errata should always be under https
At the moment, we do not check if the EDs are under https. We should probably check that.

/cc @plehegar

jennyliang220:
Fixed #906

### DIFF
--- a/lib/rules/headers/dl.js
+++ b/lib/rules/headers/dl.js
@@ -192,14 +192,14 @@ exports.check = function (sr, done) {
 
     if (dts.Implementation) {
         var linkImple = dts.Implementation.dd.querySelector("a") && dts.Implementation.dd.querySelector("a").href;
-        if (linkImple && !linkImple.match(/^https:\/\//)) {
+        if (linkImple && !linkImple.trim().toLowerCase().startsWith('https://')) {
             sr.error(self, "implelink-should-be-https", {link: linkImple});
         }
     }
 
     if (dts.EditorDraft) {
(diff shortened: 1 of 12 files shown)


lib/rules/headers/dl.js
```diff
@@ -190,6 +190,20 @@ exports.check = function (sr, done) {
         else sr.error(self, "supersedes-syntax");
     }
 
+    if (dts.Implementation) {
+        var linkImple = dts.Implementation.dd.querySelector("a") && dts.Implementation.dd.querySelector("a").href;
+        if (linkImple && !linkImple.match(/^https:\/\//)) {
```
deniak, line 195: if you only want to check if a string starts with another string, you can use [`startsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith). It's also best to [`lowercase`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLowerCase) the string first to make sure you test strings like 'Https://'...

The comment also applies in the other changes.

lib/rules/sotd/implementation.js
```diff
@@ -23,6 +23,9 @@ exports.check = function (sr, done) {
 
           for (item in url_candidates) {
             sr.info(self, 'candidate', {text: url_candidates[item], url: item});
+            if (!item.match(/^https:\/\//)) {
+              sr.error(self, 'link-should-be-https', { link: item });
```
deniak, line 27: You probably don't want to add the error in the `for` loop. If we have find 10 links, we only want 1 error.

PR #974, opened 2020-09-25
Add link check

### LINKED ISSUE #758: rule out w3c-test.org
We shouldn't have links in spec headers linking to w3c-test.org this is un unstable server and shouldn't cannot be relied on.
@sideshowbarker , feel to pitch if you think otherwise.

jennyliang220:
Fixed #758

### DIFF
--- a/lib/rules/links/reliability.js
+++ b/lib/rules/links/reliability.js
@@ -8,9 +8,8 @@ const unreliableLinks = ['w3c-test.org'];
 
 exports.check = function (sr, done) {
     sr.jsDocument.querySelectorAll("a").forEach(function (element) {
-        var href = element.href;
-        var splitLink = href.startsWith('http') && href.split('/');
-        var domain = splitLink && splitLink[2];
+        const url = new URL(element.href);
+        const domain = url.hostname;
         if (domain) {
             if (unreliableDomains.includes(domain)) {
                 const text= sr.norm(element.textContent);
--- a/lib/rules/links/reliability.js
+++ b/lib/rules/links/reliability.js
@@ -12,12 +12,10 @@ exports.check = function (sr, done) {
         var splitLink = href.startsWith('http') && href.split('/');
         var domain = splitLink && splitLink[2];
         if (domain) {
-            unreliableLinks.forEach(unreliableLink => {
-                if (domain.match(unreliableLink + '$')) {
-                    var text = sr.norm(element.textContent);
-                    sr.warning(self, "unreliable-link", { link: unreliableLink, text: text });
-                }
-            });
+            if (unreliableDomains.includes(domain)) {
+                const text= sr.norm(element.textContent);
+                sr.warning(self, "unreliable-link", { link: domain, text: text });
+            }
         }
     });
     done();
--- a/lib/rules/links/reliability.js
+++ b/lib/rules/links/reliability.js
@@ -8,8 +8,14 @@ const unreliableDomains = ['w3c-test.org', 'www.w3c-test.org'];
 
 exports.check = function (sr, done) {
     sr.jsDocument.querySelectorAll("a").forEach(function (element) {
-        const url = new URL(element.href);
-        const domain = url.hostname;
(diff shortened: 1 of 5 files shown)


lib/rules/links/reliability.js
```diff
@@ -0,0 +1,24 @@
+const self = {
+    name: 'links.reliability'
+,   section: 'document-body'
+};
+
+exports.name = self.name;
+const unreliableLinks = ['w3c-test.org'];
```
deniak, line 7:
```suggestion
const unreliableDomains = ['w3c-test.org', 'www.w3c-test.org'];
```

lib/rules/links/reliability.js
```diff
@@ -0,0 +1,24 @@
...
+
+exports.check = function (sr, done) {
+    sr.jsDocument.querySelectorAll("a").forEach(function (element) {
+        var href = element.href;
+        var splitLink = href.startsWith('http') && href.split('/');
+        var domain = splitLink && splitLink[2];
```
deniak, line 13: It's best to use the [`url` module](https://nodejs.org/api/url.html) to get the hostname.

```suggestion
        const url = new URL(element.href);
        const domain = url.hostname;
```

lib/rules/links/reliability.js
```diff
@@ -0,0 +1,24 @@
...
+        if (domain) {
+            unreliableLinks.forEach(unreliableLink => {
+                if (domain.match(unreliableLink + '$')) {
+                    var text = sr.norm(element.textContent);
+                    sr.warning(self, "unreliable-link", { link: unreliableLink, text: text });
+                }
+            });
```
deniak, line 20:
```suggestion
            if (unreliableDomains.includes(domain)) {
                const text= sr.norm(element.textContent);
                sr.warning(self, "unreliable-link", { link: domain, text: text });
            }
```

deniak (changes requested):

lib/rules/links/reliability.js
```diff
@@ -0,0 +1,21 @@
...
+exports.name = self.name;
+const unreliableDomains = ['w3c-test.org', 'www.w3c-test.org'];
+
+exports.check = function (sr, done) {
+    sr.jsDocument.querySelectorAll("a").forEach(function (element) {
+        const url = new URL(element.href);
+        const domain = url.hostname;
+        if (domain) {
+            if (unreliableDomains.includes(domain)) {
+                const text= sr.norm(element.textContent);
```
jennyliang220, line 15:
```suggestion
                const text = sr.norm(element.textContent);
```

deniak (dismissed):

lib/rules/links/reliability.js
```diff
@@ -0,0 +1,21 @@
...
+    name: 'links.reliability'
+,   section: 'document-body'
+};
+
+exports.name = self.name;
+const unreliableDomains = ['w3c-test.org', 'www.w3c-test.org'];
+
+exports.check = function (sr, done) {
+    sr.jsDocument.querySelectorAll("a").forEach(function (element) {
+        const url = new URL(element.href);
```
deniak, line 11: There's an issue with relative URLs. See https://nodejs.org/api/url.html#url_new_url_input_base

```suggestion
        const url = new URL(element.href, sr.url);
```

deniak (changes requested):

lib/rules/links/reliability.js
jennyliang220: When checking a local file (`validate?file=`) in test:
https://github.com/w3c/specberus/blob/c643b1afe879f1ef8b9e28f78431a3620d8a1961/test/api.js#L119

sr.url is `null` because it's cleared:
https://github.com/w3c/specberus/blob/c643b1afe879f1ef8b9e28f78431a3620d8a1961/lib/validator.js#L26

which still breaks `new URL()`. So I add a try catch and done() in catch, to not check the domain here.

deniak (approved):

PR #983, opened 2020-10-02
chore: Add GitHub Action Dependabot bumping

nschonni:
Will bump GitHub Actions version like `actions/setup-node` to later versions on the same schedule as the other Dependabot updates


deniak (approved):
Thanks!

PR #987, opened 2020-10-02
chore: Setup Coveralls in CI

nschonni:
Alternately, this could be split out to a separate job that only runs if affected files are touched

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -14,15 +14,31 @@ jobs:
         node-version: [12.x, 14.x]
     steps:
     - uses: actions/checkout@v2
+
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v2.1.2
       with:
         node-version: ${{ matrix.node-version }}
+
     - run: npm install
-    - run: npm run build --if-present
+    
+    - run: npm run lint
+
+    - run: npm run hint
+
+    - run: npm run jsdoc
+
+    - run: npm run coverage
+      if: secrets.W3C_API_KEY != ""
       env:
         W3C_API_KEY: ${{ secrets.W3C_API_KEY }}
         SKIP_NETWORK: true 
+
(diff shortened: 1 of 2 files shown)


deniak (approved):

.github/workflows/node.js.yml
```diff
@@ -14,15 +14,31 @@ jobs:
...
-    - run: npm run build --if-present
+    
+    - run: npm run lint
+
+    - run: npm run hint
+
+    - run: npm run jsdoc
+
+    - run: npm run coverage
+      if: secrets.W3C_API_KEY != ""
```
nschonni, line 32: Sorry, it doesn't look like it likes the syntax. I'll try and send a quick fix if you don't beat me to it

deniak: That's odd as the [doc](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif) suggests we can omit the `${{ }}`.

nschonni: Sent a fix in the new PR. It seems like actions are just bugged when you try to use `secrets` directly in a conditional. Found a way through hoisting the `env` globally and testing it instead

nschonni: PS: looks like coveralls still thinks `master` is your default branch https://coveralls.io/github/w3c/specberus

deniak: Yes, fixed now. Thanks for noticing it and thanks for improving the action workflow. Note that we may need to revise the strategy in the near future as we want to run the tests for each PR. The fact that it requires an API key that is secret doesn't help but we'll think about a new strategy.

nschonni: No worries, the `pull_request_target` will let PRs access the secret. This guarding was more for folks working off a fork and not wanting to set on up for small things

PR #992, opened 2020-10-14
Throw error when doc is neither TR nor MEM-SUBM

### LINKED ISSUE #978: Only check TR documents and submissions

jennyliang220:
Closes #978


jennyliang220:
https://github.com/w3c/specberus/blob/687cf10f9649900d9d8ad61bae858dc460b260a6/lib/validator.js#L58-L71

All checks in metadata are triggered together. So if we detect the input document is not TR(nor SUBM), that's when profile.js is done, we cannot stop other checks.

deniak (approved):

PR #999, opened 2020-10-20
Add error detail for API result

### LINKED ISSUE #706: Add an option to return the actual error values

jennyliang220:
Closes #706

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -7,6 +7,7 @@ const self = require('../package')
 ,   sink = require('./sink')
 ,   util = require('./util')
 ,   validator = require('./validator')
+,   l10n = require('./l10n')
 ;
 
 const Sink = sink.Sink
@@ -97,8 +98,10 @@ const processRequest = function(req, res) {
                     warnings2 = [];
                     info2 = [];
                     v2 = new validator.Specberus();
-                    handler2 = new Sink(function(...data2) {
-                        errors2.push(Object.assign({}, ...data2));
+                    handler2 = new Sink(function (...data2) {
+                        var data = Object.assign({}, ...data2);
+                        data.message = l10n.assembleData(null, data.name, data.key, data.extra).message;
+                        errors2.push(Object.assign({}, data));
                     }, function() {
                         sendJSONresult(errors2, warnings2, info2, res, meta);
                     }, function(...data2) {
--- a/lib/api.js
+++ b/lib/api.js
@@ -37,7 +37,7 @@ const sendJSONresult = function(err, warn, inf, res, metadata) {
 
 const getErrorMessage = function (...data2) {
     var data = Object.assign({}, ...data2);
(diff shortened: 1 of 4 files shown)


lib/api.js
```diff
@@ -97,14 +107,14 @@ const processRequest = function(req, res) {
                     warnings2 = [];
                     info2 = [];
                     v2 = new validator.Specberus();
-                    handler2 = new Sink(function(...data2) {
-                        errors2.push(Object.assign({}, ...data2));
+                    handler2 = new Sink(function (...data2) {
```
deniak, line 110: This is not related to your changes but `handler2` should also listen to `exception` events like `handle` does on l.123

deniak (changes requested):
This currently only works with the `auto` profile. You also need to add the `getErrorMessage` in the other functions composing the handler at l.134

deniak (changes requested):
The API part looks good but as mentioned during our last call, the error message should also be returned in the `validate` function

lib/util.js
```diff
@@ -18,6 +20,9 @@ const files1 = fs.readdirSync(`${__dirname}/profiles/TR/`),
  */
 
 const buildJSONresult = function(err, warn, inf, metadata) {
+    err.every(insertDetailMessage);
```
deniak, line 23: The issue with adding the `detailMessage` in `buildJSONresult` is that it's only available when pubrules is done processing all the rules (see https://github.com/w3c/specberus/blob/main/lib/validator.js#L69). It would make more sense to add the message on each error/warning/info detected (on validator.js from lines [L.48](https://github.com/w3c/specberus/blob/main/lib/validator.js#L48) to [L.50](https://github.com/w3c/specberus/blob/main/lib/validator.js#L50) and [L.106](https://github.com/w3c/specberus/blob/main/lib/validator.js#L106) to [L.108](https://github.com/w3c/specberus/blob/main/lib/validator.js#L108)

deniak (changes requested):

lib/util.js
jennyliang220: I add the message in [validator.js](https://github.com/w3c/specberus/pull/999/files#diff-75a677c2b40d5af505ce7b663015040bc3ff16953b75b43e44ddebe88c59d836) for `Specberus.prototype.error`, when emitting the error/warning.

```
this.sink.emit('warning', rule, {
    key: key,
    extra: extra,
    detailMessage: l10n.assembleData(null, rule, key, extra).message
});
```
So in https://github.com/w3c/specberus/blob/main/lib/validator.js#L107 you'll have it in `data` already.
```
self.sink.on('warning', function (...data) { warnings.push(Object.assign({}, ...data)); });
```

deniak (approved):

deniak:
Thanks @jennyliang220. Merging now.

PR #1004, opened 2020-10-21
Allow linking to a specific label in issues as well.

### LINKED ISSUE #1003: Bug in rule “headers.mailing-list”: [No mailing list found]
This bug used to exist, was fixed, and seems to have broken again. The link to GitHub is not recognized.

 I think this is because the checker assumes `one repo per spec` and CSSWG has one repo for all specs and uses GitHub labels.

And it sees this

```
<dt>Issue Tracking:
     <dd><a href="#issues-index">Inline In Spec</a>
     <dd><a href="https://github.com/w3c/csswg-drafts/labels/css-highlight-api-1">GitHub Issues</a>
```

Mind you there is also this, which seems to be the expected form

```
<p><a href="https://github.com/w3c/csswg-drafts/issues">GitHub Issues</a> are preferred for discussion of this specification.
    When filing an issue, please put the text “css-highlight-api” in the title,
    preferably like this:
    “[css-highlight-api] <i data-lt>…summary of comment…</i>”.
```

Found [while checking `https://www.w3.org/TR/2020/WD-css-highlight-api-1-20201022/`](https://www.w3.org/pubrules/?url=https%3A%2F%2Fwww.w3.org%2FTR%2F2020%2FWD-css-highlight-api-1-20201022%2F&profile=FPWD&validation=simple-validation&noRecTrack=false&informativeOnly=false&echidnaReady=false&patentPolicy=pp2004).

tabatkins:
Closes #1003

### DIFF
--- a/lib/rules/headers/mailing-list.js
+++ b/lib/rules/headers/mailing-list.js
@@ -30,7 +30,7 @@ exports.check = function (sr, done) {
                 foundArch = true;
                 return;
             }
-            if (/^https:\/\/github.com\/[\w-]+\/[\w-]+\/(issues|labels)/.test(href)) {
+            if (/^https:\/\/github.com\/[\w-]+\/[\w-]+\/(issues|labels\/[\w-]+)/.test(href)) {
                 foundRepo = true;
                 return;
             }
--- a/lib/rules/headers/mailing-list.js
+++ b/lib/rules/headers/mailing-list.js
@@ -30,7 +30,7 @@ exports.check = function (sr, done) {
                 foundArch = true;
                 return;
             }
-            if (/^https:\/\/github.com\/[\w-]+\/[\w-]+\/issues/.test(href)) {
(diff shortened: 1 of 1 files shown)


lib/rules/headers/mailing-list.js
```diff
@@ -30,7 +30,7 @@ exports.check = function (sr, done) {
                 foundArch = true;
                 return;
             }
-            if (/^https:\/\/github.com\/[\w-]+\/[\w-]+\/issues/.test(href)) {
+            if (/^https:\/\/github.com\/[\w-]+\/[\w-]+\/(issues|labels)/.test(href)) {
```
deniak, line 33: We would need to enforce the label also:

```suggestion
            if (/^https:\/\/github.com\/[\w-]+\/[\w-]+\/(issues|labels\/[\w-]+)/.test(href)) {
```

deniak (changes requested):

deniak (approved):

PR #1018, opened 2020-11-04
Integrat "Snapshot" and "Draft" in specification subtitle

### LINKED ISSUE #1006: [CR] Integrating "Snapshot" and "Draft" in specification subtitle
Request from @fantasai:
```
<fantasai> while we're on the topic, I think it would be nice to fold the "Draft" and "Snapshot" words that are currently in a special <h2> into the main status text
<fantasai> that would streamline things I think, and be less awkward as an HTML structure
<fantasai> do you think that's possible sometime?
<denis> yes, that's actually one option we considered while working on the mockups. We didn't want to make distinctions between the types of CRs but that's doable
<denis> Draft and Snapshot are actually in a <h3>
<fantasai> sorry, that's what I meant
<fantasai> so it would be <h2>W3C Candidate Recommendation Draft, [date]<h2>
<fantasai> <h2>W3C Candidate Recommendation Snapshot, [date]</h2>
<fantasai> rather than having yet another heading for the sub-status
<fantasai> That would look a lot cleaner, and probably read better on screenreaders, too
```

@plehegar I remember we discussed that possibility while working on the mockups. That change will require a bit of work from respec, bikeshed and pubrules but it is doable. WDYT?

jennyliang220:
Closes #1006


deniak (changes requested):
`https://github.com/w3c/specberus/blob/main/lib/rules/headers/h2-status.js#L23` also needs to be modified. Currently, the tests pass because we are only checking that the `<h2>` *starts with* the long status and for CRD and CRS, the long status is 'Candidate Recommendation'. It needs to be more accurate.

Also, the documentation (https://github.com/w3c/specberus/blob/main/lib/rules.json#L768 and https://github.com/w3c/specberus/blob/main/lib/rules.json#L875) needs to be updated.

lib/rules/metadata/profile.js
```diff
@@ -98,11 +98,9 @@ exports.check = function(sr, done) {
     };
     var checkRecType = function (sr) {
         var h2Element = sr.jsDocument.querySelector('body div.head h2');
-        var isCR = h2Element && h2Element.textContent.indexOf("Candidate Recommendation") > 0;
 
-        var h3Element = sr.jsDocument.querySelector('body div.head h2 + h3');
-        if (isCR) {
-            return (h3Element && h3Element.textContent === "Draft") ? "CRD" : "CR";
+        if (h2Element && h2Element.textContent.indexOf("Candidate Recommendation") > 0) {
+            return h2Element.textContent.indexOf('Draft') > 0 ? "CRD" : "CR";
```
deniak, line 103:
```suggestion
        return (h2Element && h2Element.textContent.indexOf("Candidate Recommendation Draft") > 0) ? "CRD" : "CR";
```

lib/rules.json
```diff
@@ -873,7 +872,6 @@
                                 "Candidate Recommendation",
```
deniak, line 872:
```suggestion
                                "Candidate Recommendation Draft",
```

lib/rules.json
```diff
@@ -766,7 +766,6 @@
...
```
deniak, line 766:
```suggestion
                                "Candidate Recommendation Snapshot",
```

deniak (changes requested):

deniak (approved):

Issue #1015, opened 2020-10-29
Make sure IG deliverables are not on REC-track


deniak:
The documents produced by IG cannot be on REC-track.

jennyliang220:
type of document | url | is REC-track
----|----|----
IG-Note | https://www.w3.org/TR/2020/NOTE-vocab-owl-time-agg-20200707/ | false
WG (by IG) | https://www.w3.org/TR/2019/WD-mini-app-white-paper-20190912/ | false

jennyliang220:
I thought about adding check in pp.js, but is's not the best solution.

### Current code
none-REC-track sentence (shown blow in code as `noRec`) are currently checked in pp.js for WG. Adding `if` for IG could be possible. When delivered by IG, the **none-REC-track sentence** must appear.
https://github.com/w3c/specberus/blob/ff960f99900accf1e4110e49fe5e08c3187bf5da/lib/rules/sotd/pp.js#L22-L27

### drawback for WG (by IG)
But this way, if WG doesn't have the **none-REC-track sentence**, the only error pubrules shows would be **Required patent policy paragraph not found.**. Another information will also shown indication the right sentence, but it could be too long to identify it's **none-REC-track sentence** that breaks the rule.

### impossible for IG-Notes
IG-Notes aren't TR documents, thus they don't have patent-policy paragraph and pp.js is not include in the rules.

### Suggestion
Adding a separate rule for **none-REC-track sentence** seems a good idea. But for WG (by IG), we still need to make sure it's in the same sentence with patent policy.

jennyliang220:
IG-Note is sure not on rec-track, so only WG (by IG) need be be checked. Still update pp.js now

PR #1040, opened 2020-11-19
Add IG REC-track check

### LINKED ISSUE #1041: Use to ES8 await&async functions
## When to use await&async functions

```
// Promise function with requests
function requestW3cAPI(url) {
    return new Promise(resolve => {
        w3capi.fetch((err,data) =>{   // pseudo-code
            resolve(data);
        })
    });
}

// Specberus rule, call done() when finished
exports.check = async function () {
    var data = await requestW3cAPI(url);
    console.log(data); // will log result for api
    done();
}
check();
```

## ES8 Compatibility
Node14 which Specberus currently using supports ES8 well, so we can start using ES8 grammar.
https://node.green/#ES2017-features-async-functions-await

## What Need to Change
For Specberus, to pass jshint check, change `.jshintrc` config to `"esversion": 8`

https://github.com/w3c/specberus/blob/b06e394a2598aabb74e3a7f326c3e09effe170d4/.jshintrc#L2

jennyliang220:
Closes #1041 #1015

### DIFF
--- a/lib/rules/sotd/pp.js
+++ b/lib/rules/sotd/pp.js
@@ -91,7 +91,7 @@ function fetchGroupNames(deliverIds) {
             w3capi.apiKey = process.env.W3C_API_KEY;
             w3capi.group(deliverId).fetch((err, data) => {
                 if (err) {
-                    reject();
+                    reject(err);
                 } else {
                     resolve(data.name);
                 }
@@ -138,6 +138,8 @@ exports.check = async function (sr, done) {
             sr.isIGDeliverable = groupNames.some(groupName => {
                 return groupName.endsWith('Interest Group');
             });
+        }).catch(function (e) {
+            sr.error(self, "no-response", {error: JSON.stringify(e)});
         });
 
         var pp = findPP(util.filter(sotd, "p").concat(...sotd.querySelectorAll("p")), sr);
--- a/lib/rules/sotd/pp.js
+++ b/lib/rules/sotd/pp.js
@@ -123,6 +123,12 @@ exports.check = async function (sr, done) {
     ,     pp2017 = "https://www.w3.org/Consortium/Patent-Policy-20170801/"
     ,     patent = (sr.config.patentPolicy === "pp2020") ? pp2020 : (sr.config.patentPolicy === "pp2004") ? pp2017: sr.config.patentPolicy
     ;
+    var groupName = await fetchGroupName(sr.getDelivererIDs());
+    sr.isIGDeliverable = groupName.endsWith('Interest Group');
+
+    // fetchGroupName(sr.getDelivererIDs()).then(function (groupName) {
+    //     var isIG = groupName.endsWith('Interest Group');
+    // })
 
     // Skip patent policy check for publications from TAG
     if (sr.getDelivererIDs().includes(util.TAG_ID)) return done();
--- a/lib/rules/sotd/pp.js
+++ b/lib/rules/sotd/pp.js
@@ -84,21 +84,22 @@ function findPP (candidates, sr) {
     return pp;
 }
 
-function fetchGroupName(deliverIds) {
-    return new PowerPromise(function (resolve, reject) {
-        w3capi.apiKey = process.env.W3C_API_KEY;
-        w3capi.group(deliverIds[0]).fetch((err, data) => {
-            if (err) {
-                reject();
-            } else {
-                resolve(data.name);
-            }
-        });
+function fetchGroupNames(deliverIds) {
+    var promiseArray = [];
+    deliverIds.forEach(deliverId => {
+        promiseArray.push(new PowerPromise(function (resolve, reject) {
+            w3capi.apiKey = process.env.W3C_API_KEY;
+            w3capi.group(deliverId).fetch((err, data) => {
+                if (err) {
+                    reject();
+                } else {
+                    resolve(data.name);
+                }
+            });
+        }));
     });
+    return promiseArray;
 }
-
-    // deliverIds.some(deliverId => {
-    // })
 
 const self = {
     name: 'sotd.pp'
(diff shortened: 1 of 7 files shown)


lib/rules/sotd/pp.js
```diff
@@ -81,6 +84,22 @@ function findPP (candidates, sr) {
...
+            if (err) {
+                reject();
+            } else {
+                resolve(data.name);
+            }
+        });
+    });
+}
+
+    // deliverIds.some(deliverId => {
```
nschonni, line 100: Delete?

lib/rules/sotd/pp.js
```diff
@@ -103,6 +122,12 @@ exports.check = function (sr, done) {
     ,     pp2017 = "https://www.w3.org/Consortium/Patent-Policy-20170801/"
     ,     patent = (sr.config.patentPolicy === "pp2020") ? pp2020 : (sr.config.patentPolicy === "pp2004") ? pp2017: sr.config.patentPolicy
     ;
+    var groupName = await fetchGroupName(sr.getDelivererIDs());
+    sr.isIGDeliverable = groupName.endsWith('Interest Group');
+
+    // fetchGroupName(sr.getDelivererIDs()).then(function (groupName) {
```
nschonni, line 128: Delete?

lib/rules/sotd/pp.js
jennyliang220: Hi @nschonni , thanks for the review. This PR is not finished, just push to view diff and discuss. This `deliverIds` is an array and need iteration, I'll continue working on it today.

lib/rules/sotd/pp.js
jennyliang220: Same reason I leave comment here, to discuss whether use `await` or Promise `then`. Now I tend to use `await` to make the code easy to read.

lib/rules/sotd/pp.js
nschonni: Sorry about that, I missed the Draft status

jennyliang220:
Tried with `Promise.any` locally, although document is in [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any), still get `TypeError: Promise.any is not a function` when running. [Browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any#Browser_compatibility) shows it's not available until Node15 (currently specberus uses Node14)
image

nschonni:
Might be worth adding https://www.npmjs.com/package/eslint-plugin-promise

lib/rules/sotd/pp.js
```diff
@@ -110,6 +130,15 @@ exports.check = function (sr, done) {
             return done();
         }
 
+        // Use Promise.all to get Group Names, if groups include IG, will check `noRec` sentence in patent policy paragraph.
+        await Promise.all(fetchGroupNames(sr.getDelivererIDs())).then(function (groupNames) {
+            sr.isIGDeliverable = groupNames.some(groupName => {
+                return groupName.endsWith('Interest Group');
+            });
```
deniak, line 137:
```suggestion
            sr.isIGDeliverable = groupNames.some(groupName => groupName.endsWith('Interest Group'));
```

deniak (changes requested):
The rule looks good. However, I don't understand the test you updated. The test should check that a WD published by an IG triggers an error if the sentence 'The group(s) do/does not expect this document to become a W3C Recommendation'. is missing.

lib/rules/sotd/pp.js
jennyliang220: Thanks for the reminder. Update for `doc/wg-note2.html` is adding two delivererID, but it shouldn't be in this paragraph. Will move to another paragraph.

lib/rules/sotd/pp.js
```diff
@@ -21,7 +24,7 @@ function buildWanted (groups, sr) {
 
         var noRec = ((groups.length == 2) ? "The groups do not " : "The group does not ") +
             "expect this document to become a W3C Recommendation\\. ?";
-        if (config.recTrackStatus && config.noRecTrack)
+        if ((config.recTrackStatus && config.noRecTrack )|| sr.isIGDeliverable)
```
deniak, line 27: `isIGDeliverable` doesn't need to be global. It's used only in that file.

deniak (changes requested):
I'm still not getting how the test reflects what the rules is supposed to check.

lib/rules/sotd/pp.js
jennyliang220: > I'm still not getting how the test reflects what the rules is supposed to check.

In 'test/docs/headers/wg-note2.html' I add a paragraph, saying the document is delivered by '40318', '109611' two groups. The second id '109611' is a **Interest Group**, that means the sentence `noRec` should be in PP paragraph.

The paragraph pp.js rule is looking for:
> /^This document was produced by a group operating under the( 15 September 2020| 1 August 2017)? W3C Patent Policy\. ?The group does not expect this document to become a W3C Recommendation\. ?$/

But in 'test/docs/headers/wg-note2.html' there is no such paragraph. So in 'test/rules.js' I add error "sotd.pp.no-pp" when checking 'wg-note2.html'.

deniak: > But in 'test/docs/headers/wg-note2.html' there is no such paragraph. So in 'test/rules.js' I add error "sotd.pp.no-pp" when checking 'wg-note2.html'.

The sentence is actually in that document but in a separate paragraph. The current test is confusing. I believe wg-note2.html was create to make sure both sentences are in the same paragraph.
Now, it's also mixing what we expect from an IG so in the end, we don't really know what we are testing.

PR #1053, opened 2020-11-30
chore: Add ESLint Prettier setup and Drop JSHint

nschonni:
I don't always agree with the prettier choices, but it gives a good consistent style to a codebase.
Found that the initial cleanup caused conflicts with JSHint. Since you're already using ESLint, I think it's OK to just drop JSHint as it is redundant to run both.

Feel free to close if you don't like the style that Prettier gives, as the drawback is that there are limited customization for their style. The only setting I overrode was for the single quotes, since it seemed to be used a little more.

PS: You can now see the JSDoc ESLint warnings on the files tab with the new Lint JS job


jennyliang220:
I agree dropping JSHint, since ESLint already covers the code style part. As for this PR, it changed lot of files and might take some time to review.

jennyliang220:
Hi @nschonni, it's a good idea to use eslint plugins like `prettier`, but there are different kind of code style and more investigation is needed. So I'll just leave this PR open until one code style is selected (very soon) to fit all the applications.

nschonni:
@jennyliang220 no problem at all

Issue #471, opened 2016-07-21
Headers links: check href===textNode


deniak:
Make sure the links in the headers are consistent with the text nodes.

jennyliang220:
Should `Participate:` be included? 

e.g.

```
<dt>Participate:</dt>
<dd>
    <a href="https://github.com/w3c/mediacapture-main/">GitHub w3c/mediacapture-main</a>
</dd>
```

PR #1067, opened 2020-12-11
Show more information when shortname or link differ

jennyliang220:


jennyliang220:
`dts.This.dd.querySelector("a");` and `dts.Latest.dd.querySelector("a");` and more similar paragraphs can be merged into one function.

deniak (changes requested):
Looks good but as discussed over the phone, it would be good to also tackle #471 at the same time for the implementation report and editor draft.

deniak (approved):

PR #1078, opened 2020-12-18
It's #proposed-corrections and not #proposed-correction

plehegar:
Fixing the link
 https://www.w3.org/2020/Process-20200915/#proposed-correction
to be
 https://www.w3.org/2020/Process-20200915/#proposed-corrections

And yes, it's still
 https://www.w3.org/2020/Process-20200915/#proposed-addition
so no change there.

 Relates to https://github.com/w3c/csswg-drafts/issues/5807


jennyliang220:
The change for documentation in lib/rules.json LGTM. But the rule in sotd/publish.js also need to be updated. I'll update in later commit.

test/docs/sotd/rec-publish-corrections-fail2.html
```diff
@@ -0,0 +1,34 @@
...
+            <time class="dt-published" datetime="2020-09-15">15 September 2020</time>
+        </h2>
+    </div>
+    <section id="sotd" data-link-for="Performance" class="introductory">
+        <h2>Status of This Document</h2>
+        <!-- fail check sotd.publish with error "url-text-not-found" for having wrong text for <a>. The text should include words "proposed corrections" not "includes proposed" -->
+        <p>
+            This document was published by the <a href="https://www.w3.org/webperf/">Web Performance Working Group</a> as a Recommendation.
+            <!-- substantiveChanges: true, newFeatures: false -->
+            It <a href="https://www.w3.org/2020/Process-20200915/#proposed-corrections">includes proposed</a> corrections.
```
deniak, line 26: It would probably be better to test the case @plehegar  is trying to fix

```suggestion
            It includes <a href="https://www.w3.org/2020/Process-20200915/#proposed-corrections">proposed correction</a>.
```

deniak (changes requested):

test/docs/sotd/rec-publish-corrections-fail2.html
deniak: plural will actually passed

PR #1086, opened 2020-12-29
Remove draft-stability check for Notes

### LINKED ISSUE #1069: Pubrules for WG Note seems off
https://www.w3.org/pubrules/doc/rules/?profile=WG-NOTE

There's some requirements here that seem to make no sense.

- “This is a draft document and may be updated, replaced or obsoleted by other documents at any time.” WG Notes are not draft documents any more than CRs and RECs are.

-  “This document was produced by a group operating under the W3C Patent Policy.” The Patent Policy does not apply to NOTEs, only to REC-track documents.

jennyliang220:
resolves #1069

### DIFF
--- a/lib/profiles/TR/WG-NOTE.js
+++ b/lib/profiles/TR/WG-NOTE.js
@@ -11,7 +11,7 @@ exports.config = {
 
 const base = require("../base");
 
-exports.rules = base.extendWithInserts({
+const rules = base.extendWithInserts({
     "headers.h2-status":    [
                                 require("../../rules/headers/mailing-list")
                             ,   require("../../rules/headers/copyright")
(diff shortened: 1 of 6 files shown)


lib/profiles/TR/WG-NOTE.js
```diff
@@ -11,7 +11,7 @@ exports.config = {
 
 const base = require("../base");
 
-exports.rules = base.extendWithInserts({
+const rules = base.extendWithInserts({
```
deniak, line 14: `base.RemoveRules` does reassign the first argument so if rules is a constant, there will be an exception.
```suggestion
var rules = base.extendWithInserts({
```

deniak (changes requested):
That PR doesn't resolve #1069 entirely. There's also the check on the patent policy @wseltzer agreed to drop.

jennyliang220:
Pubrules will address this in Process 2021.

PR #1152, opened 2021-03-02
Handle redirections for resources

jennyliang220:


lib/rules/links/linkchecker.js
```diff
@@ -71,14 +71,30 @@ exports.check = async function (sr, done) {
...
-            sr.error(compound, 'response-error', {
-                url,
-                status: response.status(),
-                text: response.statusText(),
-                referer,
-            });
+        // check if every resource's status code is ok, ignore 3xx
+        if (
+            !response.ok() &&
+            response.status() >= 400 &&
```
deniak, line 77: `ok()` will return true for code between 200 and 299 so this is redundant
```suggestion
            response.status() >= 400 &&
```

deniak (changes requested):

deniak (approved):

PR #1168, opened 2021-03-11
Adapt echidna, change error into warning

### LINKED ISSUE #1155: Checking compound document is not compatible with Echidna
The [linkchecker rule](https://github.com/w3c/specberus/blob/main/lib/rules/links/linkchecker.js) checks if all the resources are either in the same directory or in the allow list.
This is a problem for Echidna as it installs the document in a temporary web server to run the checks. Hence the document is not at its final location when Echidna runs pubrules.

We don't want to hardcode the echidna web server URLs in the allow list so we need a way to pass a config option so pubrules will know that the resources URLs will not start by the 'This Version' URL.

jennyliang220:
closes #1155

### DIFF
--- a/lib/l10n-en_GB.js
+++ b/lib/l10n-en_GB.js
@@ -137,11 +137,11 @@ exports.messages = {
     'links.linkchecker.display':
         'Please verify using <a href="https://validator.w3.org/checklink?uri=${link}&recursive=on">Link Checker</a>.',
     'links.linkchecker.not-same-folder':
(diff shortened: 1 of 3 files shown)


deniak (changes requested):
I'm actually not a big fan of switch errors to warnings temporarily. I would rather keep the error and deploy on a specific date (which, btw, should be announced to spec-prod@w3.org).
Otherwise, linkchecker.js looks good.

PR #1179, opened 2021-03-16
Allow document from other server, add allow list for document rule

### LINKED ISSUE #1155: Checking compound document is not compatible with Echidna

jennyliang220:
Closes #1168 , resolve #1155

### DIFF
--- a/lib/rules.json
+++ b/lib/rules.json
@@ -1868,7 +1868,7 @@
             "compound": {
                 "name": "8. Compound Documents",
                 "rules": {
-                    "compoundFilesLocation": "If the document is compound (i.e., if it consists of more than one file), all the files <span class=\"rfc2119\">must</span> be under a directory /TR/YYYY/@{param1}-shortname-YYYYMMDD/",
+                    "compoundFilesLocation": "If the document is compound (i.e., if it consists of more than one file), all the files <span class=\"rfc2119\">must</span> be under a directory /TR/YYYY/@{param1}-shortname-YYYYMMDD/. Exceptions are resources under these paths:<ol><li>https://www.w3.org/StyleSheets/TR/</li><li>https://www.w3.org/scripts/TR/</li></ol>",
                     "compoundOverview": "The main page <span class=\"rfc2119\">should</span> be called Overview.html.",
                     "compound": "All other files <span class=\"rfc2119\">must</span> be reachable by links from the document."
                 }
--- a/lib/rules.json
+++ b/lib/rules.json
@@ -1868,7 +1868,7 @@
             "compound": {
                 "name": "8. Compound Documents",
                 "rules": {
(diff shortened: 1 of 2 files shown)


lib/rules/links/linkchecker.js
```diff
@@ -14,8 +14,8 @@ const compound = {
 };
 
 const allowList = [
-    new RegExp('^https://www.w3.org/StyleSheets/TR/2016/'),
-    new RegExp('^https://www.w3.org/scripts/'),
+    new RegExp('^https://www.w3.org/StyleSheets/TR/'),
+    new RegExp('^https://www.w3.org/scripts/TR/'),
```
deniak, line 18: We can allow any scripts under `/scripts/`. We do have control over that folder.
```suggestion
    new RegExp('^https://www.w3.org/scripts/'),
```

lib/rules.json
```diff
@@ -1868,7 +1868,7 @@
             "compound": {
                 "name": "8. Compound Documents",
                 "rules": {
-                    "compoundFilesLocation": "If the document is compound (i.e., if it consists of more than one file), all the files <span class=\"rfc2119\">must</span> be under a directory /TR/YYYY/@{param1}-shortname-YYYYMMDD/",
+                    "compoundFilesLocation": "If the document is compound (i.e., if it consists of more than one file), all the files <span class=\"rfc2119\">must</span> be under a directory /TR/YYYY/@{param1}-shortname-YYYYMMDD/. Exceptions are resources under these paths:<ol><li>https://www.w3.org/StyleSheets/TR/</li><li>https://www.w3.org/scripts/TR/</li></ol>",
```
deniak, line 1871:
```suggestion
                    "compoundFilesLocation": "If the document is compound (i.e., if it consists of more than one file), all the files <span class=\"rfc2119\">must</span> be under a directory /TR/YYYY/@{param1}-shortname-YYYYMMDD/. Exceptions are resources under these paths:<ol><li>https://www.w3.org/StyleSheets/TR/</li><li>https://www.w3.org/scripts/</li></ol>",
```

lib/rules/links/linkchecker.js
```diff
@@ -47,28 +47,29 @@ exports.check = async function (sr, done) {
...
-    const thisLink = dts && dts.This && dts.This.dd.querySelector('a').href;
-    if (!thisLink) {
+    // sr.url is used as base url. Every other resources should use in same folder as base. e.g.
+    // - spec doc: https://www.w3.org/TR/2021/WD-pubrules-20210401/
+    // - image (pass): https://www.w3.org/TR/2021/WD-pubrules-20210401/images/sample.png
+    // - image (pass): https://www.w3.org/TR/2021/WD-pubrules-20210401/sample.png
+    // - image (error): https://w3c.github.io/pubrules/sample.png
+
+    if (!sr.url) {
+        await browser.close();
```
deniak, line 57: As you mentioned in our last call, you can move the launch of puppeteer after that `if`

deniak (changes requested):

deniak (approved):

Nuckijfuta (commented):
Revert

PR #1198, opened 2021-04-13
Add tools.ietf.org to unreliable service

### LINKED ISSUE #1178: Use datatracker.ietf.org URLs for IETF documents
Given
  https://mailarchive.ietf.org/arch/msg/wgchairs/CsbU8h-OEqZ4SagykQcuJkr9AZg/

we should generate warnings (or errors?) if one uses https://tools.ietf.org/.

specref.org is moving away from it:
 https://github.com/tobie/specref/pull/641

so the transition should be easy for most.

@dontcallmedom

jennyliang220:
Resolves #1178


deniak (approved):

deniak:
Let's plan a deployment tomorrow.

PR #1242, opened 2021-05-18
Give out error if the h2 profile cannot be found

### LINKED ISSUE #1209: Profile dectection error
If specberus cannot guess the profile from the `<h2>`, the error is a bit confusing:
`[EXCEPTION] The document could not be parsed, it's neither a TR document nor a Member Submission.`

If specberus cannot guess the profile because of the [recent subtitle rule update](https://github.com/w3c/specberus/pull/1201), then it should report that the id should be removed.

jennyliang220:
Resolves #1209

### DIFF
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -288,6 +288,11 @@ Specberus.prototype.getDocumentDate = function () {
     const element = this.jsDocument.querySelector(
         'body div.head h2:not(#subtitle)'
     );
+    if (!element) {
+        throw new Error(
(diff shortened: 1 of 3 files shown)


lib/validator.js
```diff
@@ -288,6 +288,11 @@ Specberus.prototype.getDocumentDate = function () {
     const element = this.jsDocument.querySelector(
         'body div.head h2:not(#subtitle)'
     );
+    if (!element) {
```
deniak, line 291: Adding the error here makes it more difficult to test with the current test suite (note that the test suite is broken with that update).
It seems more appropriate to update `lib/rules/metadata/profile.js` and `lib/rules/metadata/docDate.js` instead which are the rules used when selecting the profile 'auto'.
https://github.com/w3c/specberus/blob/main/lib/l10n-en_GB.js#L26 should also be updated to provide the same kind of details.

deniak (changes requested):

deniak (approved):

PR #1261, opened 2021-05-28
Add better error message for ED

### LINKED ISSUE #1258: Better error reporting when checking an ED
When pubrules checks an ED, it'll simply report the following error:
`Error: [EXCEPTION] The document could not be parsed, it's neither a TR document nor a Member Submission.` 

This can be confusing and pubrules could report a better error.

jennyliang220:
resolves #1258

### DIFF
--- a/lib/rules/metadata/profile.js
+++ b/lib/rules/metadata/profile.js
@@ -141,7 +141,7 @@ exports.check = async function (sr, done) {
         await getTitle(sr, (result) => {
             docTitle = result && result.title;
         });
-        if (candidate.match("Editor's Draft")) {
+        if (candidate.indexOf("editor's draft") > -1) {
             sr.throw(
                 `[EXCEPTION] The document "${docTitle}" seems to be an Editor's Draft, which is not supported.`
             );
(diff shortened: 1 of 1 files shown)


lib/rules/metadata/profile.js
```diff
@@ -136,8 +136,19 @@ exports.check = function (sr, done) {
...
     } else {
-        throw new Error(
-            "[EXCEPTION] The document could not be parsed, it's neither a TR document nor a Member Submission."
-        );
+        const getTitle = require('./title').check;
+        let docTitle;
+        await getTitle(sr, (result) => {
+            docTitle = result && result.title;
+        });
+        if (candidate.match("Editor's Draft")) {
```
deniak, line 144: `match` takes a regex as argument. The test won't work here. Also note that candidate is [lower-cased earlier](https://github.com/w3c/specberus/blob/main/lib/rules/metadata/profile.js#L28).
Instead, something like this should work:
```suggestion
        if (candidate.indexOf("editor's draft") > -1) {
```

deniak (changes requested):

deniak (approved):

PR #1292, opened 2021-06-24
Fix dependabot fail

jennyliang220:


deniak:
Looking more into the documentation, that PR won't fix the dependabot issue.
A [recent change with dependabot](https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/) makes it run the PR with read only permissions and it doesn't have access to the repository secrets. See also:
* https://github.com/dependabot/dependabot-core/issues/3253
* https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

AFAIK, there's no easy way unless we don't use the API key as a secret. We might want to switch to the event `pull_request_target` that will give access to the secret but there are other security issues involved. I'll think more about it but in the meantime, I'm closing that PR.

PR #1384, opened 2021-09-30
Add support for errata as a dd

marcoscaceres:
Add support for errata appearing as a dd in the head, as per:
https://w3c.github.io/tr-design/p2021mockup/rec-6.3.10.html

Happy to add a test, but could use some guidance 🙏 

Also, I was unsure if we want to continue to support `<p>`.

### DIFF
--- a/lib/rules/metadata/errata.js
+++ b/lib/rules/metadata/errata.js
@@ -9,7 +9,7 @@ exports.name = 'metadata.errata';
 exports.check = function (sr, done) {
     const errataRegex = /errata/i;
     const linkElement = sr.jsDocument.querySelectorAll(
-        'body div.head dl + p > a'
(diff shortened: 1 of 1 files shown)


deniak:
@marcoscaceres, there's actually a [pending PR](https://github.com/w3c/specberus/pull/1366) for the new process and the errata rule is updated [here](https://github.com/w3c/specberus/pull/1366/files#diff-1b10cd853b3ac3baceee0d750ddc5432687b766ed2d59d676adc520e2a6e4c5e).

I'm still reviewing that PR but I expect to have that PR deployed on a test server so you guys can use it to test the snapshots generated with respec and bikeshed. I'll let you know when it's ready.

marcoscaceres:
Excellent! Thanks for the update.

Issue #783, opened 2018-11-16
Make API key an option 


marcoscaceres:
For the API_KEY, it would be great if it could be passed as an option to `validate()`. In building a tool, I want to be able to do:

```
npx respec-validator --w3c-api-key=the-key --other-options index.html
```

which later calls:

```JS
async function doPubRulesValidation(file, profile = "WD") {
  const Specberus = require("specberus");
  const events = new EventSink();
  Specberus.validate({ file, profile, events, /* API_KEY_HERE*/ });
  await events.done();
}
```

jennyliang220:
Hi @marcoscaceres , is this still necessary? (The issue is 3 years old, so...)
If necessary, we can implement this.

marcoscaceres:
Like Octokit would be great... The main thing is just being able to pass the API key somehow.

PR #1386, opened 2021-10-02
Require apiKey in Specberus constructor instead of env

sidvishnoi:
Refactored code to better support Node API:
- `apiKey` is required when creating `new Specberus` instance (instead of when importing).
- Specberus instance doesn't rely on `process.env`
- `process.env.W3C_API_KEY` is required only in tests and server.
- `process.env.BASE_URI` is required only in server.
- Avoiding API breaking change by using `process.env.W3C_API_KEY` as
    default during construction.

Closes https://github.com/w3c/specberus/issues/783
Ref: Adding specberus to spec-prod GitHub action (https://github.com/w3c/spec-prod/pull/100)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ logs
 results
 
 package-lock.json
+yarn.lock
 npm-debug.log
 node_modules
 bower_components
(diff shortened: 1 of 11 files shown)


deniak (approved):

deniak:
Thanks @sidvishnoi!

PR #1409, opened 2021-11-04
Add patch for pp check

jennyliang220:


lib/l10n-en_GB.js
```diff
@@ -204,7 +204,7 @@ exports.messages = {
     'sotd.draft-stability.not-found': 'No stability paragraph found.',
     // sotd/pp
     'sotd.pp.no-pp':
-        'Required patent policy paragraph not found. Expected text related to patent policy requirements: <blockquote>${expected}</blockquote>',
+        'Required patent policy paragraph not found. "${expected}". Expected text related to patent policy requirements: <blockquote>${expected}</blockquote>',
```
deniak, line 207: I'm guessing that change is not wanted.

lib/rules/sotd/pp.js
```diff
@@ -145,13 +117,20 @@ exports.check = async function (sr, done) {
     const { TAG } = util;
     if (deliverId.indexOf(TAG.id) >= 0) return done();
 
-    // Make sure joint publication uses the same Patent Policy.
+    const groups = await sr.getDelivererGroups();
+    // Check if the document is published by WGs only. If IG/AB/TAG(s) are evolved, the document cannot be on REC-track and should have sentence meaning 'pp not apply'.
```
deniak, line 121:
```suggestion
    // Check if the document is published by WGs only. If IG/AB/TAG(s) are involved, the document cannot be on REC-track and should have a sentence meaning 'pp does not apply'.
```

lib/rules/sotd/pp.js
```diff
@@ -1,38 +1,41 @@
...
         wanted = `${start}was produced by ${
             groups.length === 2 ? 'groups ' : 'a group '
-        }operating under the( 15 September 2020| 1 August 2017)? W3C Patent Policy\\. ?`;
-
-        if (!isRecTrack || isIGDeliverable) wanted = NO_PP_TEXT;
+        }operating under the${ppText} W3C Patent Policy\\. ?`;
 
-        if (config.informativeOnly && config.track !== 'Note')
-            wanted += 'This document is informative only\\. ';
+        if (!isRecTrack || !publishedByWgOnly)
```
deniak, line 35: I don't think it's necessary to check if `publishedByWgOnly` is false here. All the documents that aren't on REC track will have that sentence.
This should also be integrated in the `if` at line 39 (`if...else`).

Also, I don't think that code works with Statements which have another sentence before the PP text. Not that this first sentence is being checked by `stability.js` so the rules are overlapping.

test/rules.js
```diff
@@ -944,6 +944,7 @@ const tests = {
                     patentPolicy: 'pp2004',
                     track: 'Note',
                 },
+                errors: ['sotd.pp.no-pp'],
```
deniak, line 947: it would be better to fix the doc instead.

deniak (changes requested):

deniak (approved):
I think the rule should be improved and refactored but for the time being. let's merge this.

PR #1410, opened 2021-11-05
Add stability check for registry track

jennyliang220:


deniak (changes requested):
stability.js is not checking the additional sentence needed in CRY and CRYD.

jennyliang220:
> stability.js is not checking the additional sentence needed in CRY and CRYD.

The implementation of this feature will be in a new PR, after https://github.com/w3c/tr-design/pull/271 is addressed.

lib/rules/sotd/draft-stability.js
```diff
@@ -8,7 +8,11 @@ const self = {
 exports.name = self.name;
 
 exports.check = function (sr, done) {
-    if (!sr.config.stabilityWarning || sr.config.stabilityWarning === 'REC')
+    if (
```
deniak, line 11: I don't know why this hasn't been removed earlier but if a check doesn't do anything for specific profiles, then the rule ought to be removed from the profile itself exactly like how you are doing it for CRY.

lib/profiles/TR/CRY.js
```diff
@@ -1,14 +1,17 @@
 exports.name = 'CRY';
 exports.config = {
     status: 'CRY',
-    longStatus: 'Candidate Registry Snapshot',
+    longStatus: 'Candidate Registry',
```
deniak, line 4: I don't understand how the tests passed with the previous longStatus. Are we even testing this?

deniak (changes requested):

deniak (approved):

PR #1434, opened 2021-11-19
Add usage check, apply text change for Registry

jennyliang220:


test/docs/p2021/2021-rec.html
```diff
@@ -87,7 +87,7 @@ <h2>Status of This Document</h2>
       W3C recommends the wide deployment of this specification as a standard for the Web.
     </p>
     <p>
-      A W3C Recommendation is a specification that, after extensive consensus-building, has received the endorsement of the
+      A W3C Recommendation is a specification that, after extensive consensus-building, is endorsed by the
```
deniak, line 90: What are you trying to test here? The paragraph for REC is not the expecting one. There's still "*the* W3C"

test/rules.js
```diff
@@ -832,6 +832,21 @@ const tests = {
...
+                config: {
+                    longStatus: 'Registry',
+                },
+            },
+            {
+                doc: 'p2021/2021-rec.html',
+                config: {
+                    longStatus: 'Registry',
+                },
+                errors: ['sotd.usage.not-found'],
```
deniak, line 847: Same question as above. What are you trying to test? Obviously, using a REC template will not work for Registry

deniak (changes requested):

test/rules.js
jennyliang220: ... to test the wording doesn't match and pubrules should reports an error

test/docs/p2021/2021-rec.html
jennyliang220: It doesn't matter in this case. This REC document is only used to fail the check

jennyliang220: Will fix this mockup text in manual check PR.

deniak: It's the only p2021 document for Recommendation and it's not even using the right wording. I also note, it's only used once in the whole test suite so I think that's an issue but I'll mark the discussion as resolved assuming the test suite gets refactored soon and I don't want to delay these fixes any longer

lib/l10n-en_GB.js
```diff
@@ -179,7 +182,7 @@ exports.messages = {
     'sotd.stability.no-stability':
         'No stability warning paragraph. Expected text: <blockquote>${expected}</blockquote>',
     'sotd.stability.no-rec-review':
-        'No indication that the Recommendation has received the endorsement of the W3C and its Members.',
+        'No indication that the Recommendation is endorsed by the W3C and its Members.',
```
deniak, line 185:
```suggestion
        'No indication that the Recommendation is endorsed by W3C and its Members.',
```

lib/rules.json
```diff
@@ -478,8 +478,8 @@
                                 "&lt;a href=\"https://www.w3.org/2021/Process-20211102/#recs-and-notes\"&gt;Registry track&lt;/a&gt;"
                             ],
                             "customParagraph": true,
-                            "deployment": "It <span class=\"rfc2119\">must</span> include the expections in terms of deployment. The recommended text is: <blockquote>W3C recommends the wide deployment of this registry.</blockquote>",
-                            "stability": "It <span class=\"rfc2119\">must</span> set expectations about the (in)stability of the document. The <span class=\"rfc2119\">recommended</span> text is: <blockquote class=\"boilerplate\"><p>A W3C Registry is a specification that, after extensive consensus-building, has received the endorsement of W3C and its Members.</p>",
+                            "usage": "It <span class=\"rfc2119\">must</span> include the expectations if terms of usage of this registry. The SOTD should include the paragraph: <blockquote>W3C recommends the wide usage of this registry.</blockquote>",
```
deniak, line 481: there was a typo in my earlier suggestion...

```suggestion
                            "usage": "It <span class=\"rfc2119\">must</span> include the expectations in terms of usage of this registry. The SOTD should include the paragraph: <blockquote>W3C recommends the wide usage of this registry.</blockquote>",
```

deniak (changes requested):

deniak (approved):

PR #1441, opened 2021-11-25
Organize test files, use express-handlebars and json to render test cases.

jennyliang220:


lib/rules/style/sheet.js
```diff
@@ -25,7 +25,7 @@ exports.check = function (sr, done) {
     const lnk = sr.jsDocument.querySelectorAll(stylesheetLinks.join(', '));
     if (!lnk.length) sr.error(missing, 'not-found', { url });
     else {
-        const siblings = util.nextAll(lnk, 'link[rel=stylesheet], style');
+        const siblings = util.nextAll(lnk[0], 'link[rel=stylesheet], style');
```
deniak, line 28: Does this mean that rule was always skipped?

jennyliang220: Yes, there's was a bug here

test/doc-views/TR/TR.js
```diff
@@ -0,0 +1,216 @@
+exports.data = {
+    body: {
```
jennyliang220, line 2: Hi denis, the `missing class on body` issue we discussed yesterday, it turns out to because of the 'key'. `body` seems to be a remaining word. If I rename it to `body1` or `bodyEle`, the class would be in HTML as expected.

deniak: Indeed, `body` is served as the placeholder in `express-handlebars`

lib/util.js
```diff
@@ -167,6 +167,18 @@ allProfiles.forEach(i => {
...
+        const valid =
+            /((TR|SUBM)\/([A-Za-z]+\/)?([A-Z][A-Z-]*[A-Z](-Echidna)?))\.js$/.exec(
+                i
+            );
+        if (valid && valid.length > 3)
+            // eslint-disable-next-line import/no-dynamic-require
+            profiles[valid[4]] = require(`./profiles/${valid[1]}`);
+    }
+);
+
```
deniak, line 181: What's the purpose of that change? It seems to be duplicating the previous lines.

test/data/TR/Recommendation/FPWD.js
```diff
@@ -0,0 +1,544 @@
...
+                errors: ['sotd.pp.no-pp2020'],
+            },
+            // TODO: find a patent neither pp2020 pp2017
+            // {
+            //     data: 'noPPLink',
+            //     config: {
+            //         patentPolicy: 'pp2021',
+            //     },
+            //     errors: ['sotd.pp.no-pp-link'],
+            // },
```
deniak, line 501: delete
```suggestion
```

test/docs/sotd/rec-obsl.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+><!DOCTYPE html>
```
deniak, line 1: That's probably a mistake from a commit.

deniak (changes requested):

deniak:
There seems to be some problems with the whole stack of `require`. I'm not able to run a local express server to generate the test documents. For reasons I haven't figured out yet, it works while running the tests but running the express server with the exact same code as in the tests produces the following error `TypeError: Cannot read properties of undefined (reading 'slice')` in `profileUtil.js`.
It seems to be related to a circular dependency and the rules aren't correctly set but so far, it's a mystery to me.

test/doc-views/layout/TR.handlebars
```diff
@@ -0,0 +1,259 @@
...
+                {{> stability}}
+            </span>
+        </p>
+
+        {{#config.isCR}}
+            <p>This Candidate Recommendation is not expected to advance to Proposed Recommendation any earlier than
+                {{#if config.isEchidna}}
+                    <span class="handlebars-data">{{sixMonthLater}}.</span>
+                {{else}}
+                    04 October 2022.</p>
```
deniak, line 198: This will fail starting from September 2022. It might be good to parameterize that date too.

test/doc-views/layout/TR.handlebars
```diff
@@ -0,0 +1,259 @@
...
+                {{/if}}
+        {{/config.isCR}}
+
+        {{#config.isCRY}}
+            <p>
+                This Candidate Registry is not expected to advance to Registry any earlier than 6 June 2022.
```
deniak, line 204: Same comment. Date should be a parameter

test/doc-views/layout/TR.handlebars
```diff
@@ -0,0 +1,259 @@
...
+            </p>
+        {{/config.isCRY}}
+
+        {{#config.isPR}}
+            <p>
+                <abbr title="World Wide Web Consortium">W3C</abbr> Members and other interested parties are invited to review the document and send comments through 15 October 2021. Advisory Committee Representatives should consult their <a href="{{sotd.acReviewLink}}">questionnaires</a>. Note that substantive technical comments were expected during the Candidate Recommendation review period that ended 13 July 2021.
```
deniak, line 210: Same comment. Date should be a parameter

test/rules.js
```diff
@@ -48,13 +52,13 @@ const equivalentArray = function (a1, a2) {
  * Assert that metadata detected in a spec is equal to the expected values.
  *
  * @param {String} url - public URL of a spec.
- * @param {String} file - name of local file containing a spec (without path and without ".html" suffix).
+ * @param {String} file - name of local file containing a spec (without pth and without ".html" suffix).
```
deniak, line 55: wrong global replace
```suggestion
 * @param {String} file - name of local file containing a spec (without path and without ".html" suffix).
```

test/rules.js
```diff
@@ -175,1064 +177,57 @@ describe('Basics', () => {
...
+app.get('/doc-views/:docType/:track/:profile', (req, res) => {
+    const { rule, type } = req.query;
+    // get data for template from json (.js)
+    const data = require(pth.join(
+        __dirname,
+        `./doc-views/${req.params.docType}/${req.params.track}/${req.params.profile}.js`
+    ));
+
+    let finalData;
+    if (type.startsWith('good')) {
```
deniak, line 205: Might be good to check the exact match. I'm guessing if we were to use `good-foobar`, there will be a reason and we will want to use the corresponding property.

test/rules.js
```diff
@@ -175,1064 +177,57 @@ describe('Basics', () => {
...
+        finalData = data[type];
+    } else {
+        if (!rule)
+            res.send(
+                '<h1>Error: please add the parameter "rule" in the URL </h1>'
+            );
+        if (!type)
```
deniak, line 212: With the `startsWith` above, that `if` is useless as there will be an exception before. This might not be a problem if we change that code.

test/rules.js
```diff
@@ -175,1064 +177,57 @@ describe('Basics', () => {
...
+});
+
+app.get('/doc-views/SUBM/MEM-SUBM', (req, res) => {
+    // get data for template from json (.js)
+    const { goodData } = require(pth.join(
+        __dirname,
+        `./doc-views/SUBM/MEM-SUBM.js`
+    ));
+
+    res.render(pth.join(__dirname, './doc-views/layout/TR'), goodData);
```
deniak, line 229: This probably needs a different layout. Submissions have different requirements.

test/rules.js
```diff
@@ -1253,115 +248,192 @@ app.get('/docs/links/image/logo-redirection-3', (req, res) => {
...
+after(done => {
+    if (server) {
+        server.close(done);
+    }
+});
+
+function buildHandler(test, done) {
+    const handler = new Sink();
+    handler.on('err', (type, data) => {
+        if (DEBUG) console.log(type, data);
```
deniak, line 266:
```suggestion
        if (DEBUG) console.error(type, data);
```

test/rules.js
```diff
@@ -1253,115 +248,192 @@ app.get('/docs/links/image/logo-redirection-3', (req, res) => {
...
-                            const options = {
-                                profile,
-                                events: handler,
-                            };
+    handler.on('warning', (type, data) => {
+        if (DEBUG) console.log('[W]', data);
+        handler.warnings.push(`${type.name}.${data.key}`);
+    });
+    handler.on('done', () => {
+        handler.done += 1;
```
deniak, line 274: `handler.done` is not used to check if all the rules are done.

deniak (changes requested):

deniak (approved):

PR #1443, opened 2021-11-25
Remove pp2002, remove Amended

jennyliang220:


deniak (changes requested):
the changes look good but you also need to update `lib/rules/headers/w3c-state.js` and `lib/rules/metadata/profile.js`

deniak (approved):

PR #1455, opened 2021-12-03
chore: Use lint-staged for husky

nschonni:
Makes the current commit hook a little faster by just running the linting and spellchecking on files being changed, and auto fix the ESLint issues

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "expect.js": "0.3",
     "husky": "7.0.4",
     "jsdoc": "3.6.7",
+    "lint-staged": "^12.1.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -39,6 +39,7 @@
     "expect.js": "0.3",
     "husky": "7.0.4",
     "jsdoc": "3.6.7",
+    "lint-staged": "^12.1.2",
```
jennyliang220, line 42:
```suggestion
    "lint-staged": "12.1.2",
```

Pubrules currently uses the exact version number for all the dependencies.

jennyliang220 (approved):

PR #1489, opened 2021-12-25
fix: relative path before BASE_URI

nschonni:
Noticed this because pretty was crashing trying to parse it. The other reference to these URIs don't have that relative path in the front, so I dropped it


jennyliang220 (approved):

deniak:
This is breaking all the links to the documentation of all profiles if `BASE_URI` is set. I'm reverting that PR. @jennyliang220 please, make sure you test before merging.

Issue #1378, opened 2021-09-23
Copyright and PP check for TAG documents.


jennyliang220:
The TAG's [charter](https://www.w3.org/2004/10/27-tag-charter.html) is operating PP2004 and it didn't show in "Patent Policy" which document license the TAG is using.

@ylafon @wseltzer any ideas?

deniak:
Duplicate #1481

deniak:
@wseltzer, the TAG and the AB are special groups that can publish to /TR. It seems there is no specific document license attached to these groups. Can you confirm they can use whatever license they want so we can fix the current rule?

wseltzer:
Thanks @deniak. I support removing the pubrules check for a specific license for TAG and AB documents.

Issue #1377, opened 2021-09-23
[Bug] TAG publication shouldn't check group charter


jennyliang220:
https://www.w3.org/pubrules/?url=https%3A%2F%2Fwww.w3.org%2FTR%2F2021%2FNOTE-design-principles-20210923%2F&profile=WG-NOTE&validation=simple-validation&noRecTrack=true&informativeOnly=true&echidnaReady=false&patentPolicy=

#### Error: charter
`The group does not have any current charter, please contact the team contact.`

Pubrules shouldn't check charter for TAG documents.

#### Error: document license
`Could not get the document license from the API. Please make sure this group has a current charter associated with at least one document license.`
The TAG's charter doesn't mention document license. also in #1378

deniak:
duplicate #1481

PR #1508, opened 2022-01-06
Skip copyright and charter check for TAG and AB

jennyliang220:
resolves #1508 , resolves https://github.com/w3c/specberus/issues/1378, and resolves https://github.com/w3c/specberus/issues/1377

### DIFF
--- a/lib/rules/headers/copyright.js
+++ b/lib/rules/headers/copyright.js
@@ -43,9 +43,6 @@ exports.name = self.name;
 exports.check = async function (sr, done) {
     const copyright = sr.jsDocument.querySelector('body div.head p.copyright');
     if (copyright) {
-        let allowBothLicense = false;
-        const chartersData = await sr.getChartersData();
-
         // Skip check if the document is only published by TAG and/or AB
         const TagID = util.TAG.id;
         const AbID = util.AB.id;
@@ -56,6 +53,8 @@ exports.check = async function (sr, done) {
         );
         if (!groupIds.length) return done();
 
+        let allowBothLicense = false;
+        const chartersData = await sr.getChartersData();
         if (!chartersData || !chartersData.length) {
             sr.error(self, 'no-data-from-API');
             return done();
--- a/lib/rules/headers/copyright.js
+++ b/lib/rules/headers/copyright.js
@@ -49,7 +49,7 @@ exports.check = async function (sr, done) {
         const deliverers = await sr.getDelivererIDs();
         // groupIds: a list of ids without TAG or AB
         const groupIds = deliverers.filter(
-            deliverer => [TagID, AbID].indexOf(deliverer) < 0
+            deliverer => ![TagID, AbID].includes(deliverer)
         );
         if (!groupIds.length) return done();
 
--- a/lib/rules/headers/copyright.js
+++ b/lib/rules/headers/copyright.js
@@ -13,6 +13,9 @@ const self = {
     section: 'front-matter',
     rule: 'copyright',
 };
+
+const util = require('../../util');
+
 // W3C Software and Document License
 const LICENSE_CS_TEXT = 'permissive document license';
 const LICENSE_CS_URL = 'https://www.w3.org/Consortium/Legal/copyright-software';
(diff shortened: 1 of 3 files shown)


lib/rules/headers/copyright.js
```diff
@@ -42,6 +45,17 @@ exports.check = async function (sr, done) {
     if (copyright) {
         let allowBothLicense = false;
         const chartersData = await sr.getChartersData();
+
+        // Skip check if the document is only published by TAG or AB
+        const TagID = util.TAG.id;
+        const AbID = util.AB.id;
+        const deliverers = await sr.getDelivererIDs();
+        if (
+            deliverers.length === 1 &&
```
deniak, line 54: What if we have a document produced by the TAG *and* the AB?

lib/rules/sotd/charter.js
```diff
@@ -7,14 +7,26 @@ const self = {
...
     if (sotd) {
-        const deliverId = await sr.getDelivererIDs();
-        if (!deliverId.length) {
+        const deliverIds = await sr.getDelivererIDs();
+
+        // Skip check if the document is only published by TAG or AB
+        const TagID = util.TAG.id;
+        const AbID = util.AB.id;
+        if (
+            deliverIds.length === 1 &&
```
deniak, line 24: What if we have a document produced by the TAG and the AB?

lib/rules/sotd/charter.js
```diff
@@ -7,14 +7,26 @@ const self = {
...
+            (deliverIds[0] === TagID || deliverIds[0] === AbID)
+        )
+            return done();
+
+        if (!deliverIds.length) {
```
deniak, line 29: That `if` should be the first one (before even checking if it's the TAG or the AB.

deniak (changes requested):

lib/rules/headers/copyright.js
```diff
@@ -42,6 +45,17 @@ exports.check = async function (sr, done) {
     if (copyright) {
         let allowBothLicense = false;
         const chartersData = await sr.getChartersData();
```
deniak, line 47: This should be moved after the TAG/AB check. There's no need to make additional requests to get the charters if the document is produced by the TAG/AB

lib/rules/headers/copyright.js
```diff
@@ -42,6 +45,17 @@ exports.check = async function (sr, done) {
...
+
+        // Skip check if the document is only published by TAG and/or AB
+        const TagID = util.TAG.id;
+        const AbID = util.AB.id;
+        const deliverers = await sr.getDelivererIDs();
+        // groupIds: a list of ids without TAG or AB
+        const groupIds = deliverers.filter(
+            deliverer => [TagID, AbID].indexOf(deliverer) < 0
+        );
+        if (!groupIds.length) return done();
```
deniak, line 57: The check is wrong and doesn't match the comment l.49.
If a document is a joint publication between the TAG (or AB) and a WG, then, we should look for the license associated with the WG.

lib/rules/sotd/charter.js
```diff
@@ -7,18 +7,30 @@ const self = {
...
         }
 
+        // Skip check if the document is only published by TAG and/or AB
+        const TagID = util.TAG.id;
+        const AbID = util.AB.id;
+        // groupIds: a list of ids without TAG or AB
+        const groupIds = deliverIds.filter(
+            deliverer => [TagID, AbID].indexOf(deliverer) < 0
+        );
+        if (!groupIds.length) return done();
```
deniak, line 32: As above, we should consider the case of a joint publication between the TAG (or AB) and a WG

deniak (changes requested):

lib/rules/headers/copyright.js
jennyliang220: Let's take an example:
```
const groupIds = [TagID, AbID, WG_ID].filter(
    deliverer => [TagID, AbID].indexOf(deliverer) < 0
);
````

groupIds after `filter` would be 'WG_ID', which means L57 shouldn't execute.

---
> we should look for the license associated with the WG

The license is related to the charter. And the charter's data comes from `getChartersData()`, which will also skip TAG and AB, only return information regarding the `WG`

deniak: Ok, I got confused bu the `indexOf`. Note that `includes` will be more accurate than `indexOf` because we are checking exact IDs.

PR #1509, opened 2022-01-06
Separate pp

### LINKED ISSUE #1470: Patent Policy rule can be simplefied?
Just to complain the pp.js rule is so hard to understand, both for developers and team contacts, as well as its document in rules.json. 

Maybe it needs a refactor, or split into several rules. E.g. rec-track has sentence saying `pp applies`, while note-track and registry-track saying `pp doesn't applies`.

related to https://github.com/w3c/specberus/issues/1466.

jennyliang220:
try to resolve #1470


jennyliang220:
The change doesn't make pp simpler, so I'm closing this. Maybe come back after the new test is launched.

PR #1579, opened 2022-04-13
Add check for lowercase new shortname

### LINKED ISSUE #1578: Use lowercase in new /TR shortnames
Require **new** /TR shortnames to use lowercase only from now on. For more [information](https://lists.w3.org/Archives/Public/spec-prod/2022JanMar/0048.html).

jennyliang220:
resolves #1578 .

For shortname change document, as well as first-public documents (FPWD, some DNOTE, some DRY), check the shortname doesn't contain uppercase letters.


lib/rules.json
```diff
@@ -1580,7 +1589,7 @@
...
-                    "docIDThisVersion": "The syntax of a “this version” URI <span class=\"rfc2119\">must</span> be <code>https://www.w3.org/TR/YYYY/@{param1}-shortname-YYYYMMDD/</code>.",
+                    "docIDThisVersion": "The syntax of a “this version” URI <span class=\"rfc2119\">must</span> be <code>https://www.w3.org/TR/YYYY/@{param1}-shortname-YYYYMMDD/</code>.@{param2}",
```
deniak, line 1592: To be on the safe side, I think the sentence regarding the shortname case should be there in all profiles so this also includes potential shortname changes

lib/rules/headers/dl.js
```diff
@@ -129,6 +129,16 @@ export async function check(sr, done) {
                     )
                         sr.error(self, 'this-date');
                 } else sr.warning(self, 'no-date');
+
+                // Require new /TR shortnames to use lowercase. No RY published yet, so this rule applies to all RYs.
+                const needLowercase =
```
deniak, line 134: You probably don't need to check each status. Since you are already relying on `isFP`, the only cases that would require the check are:
1. `isFP() === true`
2. shortname change (see how `isFP` is checking for such change)

deniak (changes requested):

deniak (approved):

PR #1710, opened 2023-02-27
Add rule to detect echidna deliverer change

jennyliang220:
Add rule deliverer-change to echidna files, resolves https://github.com/w3c/echidna/issues/581

- Use W3C API to get previous version's deliverer id

### DIFF
--- a/lib/rules/echidna/deliverer-change.js
+++ b/lib/rules/echidna/deliverer-change.js
@@ -10,7 +10,7 @@ const self = {
 
 export const { name } = self;
 
-async function getPreviousDelivererIDs(shortname, previousUrl) {
+function getPreviousDelivererIDs(shortname, previousUrl) {
     const date = previousUrl.match(/-(\d{8})\/$/);
     if (date)
         return new Promise(resolve => {
--- a/lib/rules/echidna/deliverer-change.js
+++ b/lib/rules/echidna/deliverer-change.js
@@ -11,17 +11,19 @@ const self = {
 export const { name } = self;
 
 async function getPreviousIDs(shortname, previousUrl) {
-    const date = previousUrl.match(/\d{8}/);
-    return new Promise(resolve => {
-        // e.g. https://api.w3.org/specifications/WGSL/versions/20230221/deliverers
-        w3cApi
-            .specification(shortname)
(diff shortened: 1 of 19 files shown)


lib/rules/echidna/deliverer-change.js
```diff
@@ -0,0 +1,57 @@
...
+    name: 'echidna.deliverer-change',
+    section: 'metadata',
+    rule: 'delivererID',
+};
+
+export const { name } = self;
+
+async function getPreviousIDs(shortname, previousUrl) {
+    const date = previousUrl.match(/\d{8}/);
+    return new Promise(resolve => {
```
deniak, line 15: The promise should reject if `date` is null or returns more than one string.

lib/rules/echidna/deliverer-change.js
```diff
@@ -0,0 +1,57 @@
...
+
+const self = {
+    name: 'echidna.deliverer-change',
+    section: 'metadata',
+    rule: 'delivererID',
+};
+
+export const { name } = self;
+
+async function getPreviousIDs(shortname, previousUrl) {
```
deniak, line 13: To avoid confusion, I suggest to rename that function. Also, `async` doesn't seem necessary here:
```suggestion
function getPreviousDelivererIDs(shortname, previousUrl) {
```

deniak (changes requested):

lib/rules/echidna/deliverer-change.js
```diff
@@ -0,0 +1,59 @@
...
+            // e.g. https://api.w3.org/specifications/WGSL/versions/20230221/deliverers
+            w3cApi
+                .specification(shortname)
+                .version(date[1])
+                .deliverers()
+                .fetch({ embed: true }, (err, data) => {
+                    resolve(data && data.map(obj => obj.id));
+                });
+        });
+    return [];
```
deniak, line 26: This is wrong. In one case the function returns a Promise and in the other, it returns an empty array...
That function should return a promise for each case...

deniak (changes requested):

deniak:
Thanks @jennyliang220

PR #1900, opened 2024-12-16
add exception on implementation report for webgpu and WGSL

deniak:
webgpu and WGSL are transitioning to CR but the implementation report will be published later on.
That PR adds an exception to pubrules to allow subsequent CRD publications.
I will remove the exception once the implementation report is ready.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -91,11 +91,13 @@
         "vcard",
         "WHATWG",
         "wcag",
+        "webgpu",
         "webidl",
         "webperf",
         "webrtc",
         "Whut",
-        "wrongprocess"
+        "wrongprocess",
+        "WGSL"
     ],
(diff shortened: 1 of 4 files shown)


tidoust (approved):

plehegar:
SGTM.

The Process requires:
[[
must document how [adequate implementation experience](https://www.w3.org/policies/process/#adequate-implementation) will be demonstrated,
]]

which we generally fulfill by linking to a draft implementation report.

The document makes it clear about its intent:
[[
The group expects to demonstrate implementation of each feature in at least two deployed browsers on top of modern GPU system APIs. The test suite will be used to build an implementation report.
]]

plehegar (approved):

deniak:
Thanks both for the review.

PR #1905, opened 2025-01-08
Bikeshed adds width and height attributes to the security-privacy-questionnaire copyright image.

jyasskin:
Ideally pubrules would ignore these attributes when comparing the copyright string, but hopefully this works around the issue.


hober (approved):
Looks good to me.

deniak (approved):

PR #1973, opened 2025-06-19
restore profiles export

deniak:
https://github.com/w3c/specberus/pull/1441 removed the `profiles` export but spec-prod still relies on it.
That PR restores it.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -53,6 +53,22 @@ export const allProfiles = [
     ...filesSUBM.map(x => `SUBM/${x}`),
 ];
 
+export const profiles = Object.fromEntries(
+    allProfiles
+        .map(file => {
+            const match =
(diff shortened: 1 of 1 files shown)


lib/util.js
```diff
@@ -53,6 +53,22 @@ export const allProfiles = [
...
+export const profiles = Object.fromEntries(
+    allProfiles
+        .map(file => {
+            const match =
+                /((TR|SUBM)\/([A-Za-z]+\/)?([A-Z][A-Z-]*[A-Z](-Echidna)?))\.js$/.exec(
+                    file
+                );
+            if (match && match[4]) {
+                const key = match[4];
+                return [key, import(`./profiles/${match[0]}`)];
```
sidvishnoi, line 65: Not sure if it should be `match[0]` or `match[1]`, but rest looks good! `match[0]` is same as `file`, right?

PR #2004, opened 2025-10-06
Allow former editors to be listed and skip the check on their group's participation

### LINKED ISSUE #2003: Relax check on "An editor must be a participant"
Via https://github.com/w3c/process/issues/948

Following #1786, Pubrules checks that editors listed under "Editors" participate in the Working Group to meet the Process requirement that ["An editor must be a participant"](https://www.w3.org/policies/process/#publication). That check assumes that the notion of editors in the Process equates the "Editors" label in the spec. In practice, groups may want to preserve a person who no longer edits a spec in the "Editors" list, e.g., because they feel that the "Former editors" list devalues the editor's status.

Conclusion from the Process issue discussion is that this is not a Process issue, and up to Pubrules. I propose to relax the check in Pubrules to allow groups to mix current and former editors under "Editors" (as we used to do).

For copyright and IPR reasons, it still seems important to identify active editors at the time of publication and make sure that they are in the group. Check in Pubrules could skip entries in the "Editors" list that have a `former` class. To make sure that the spec conveys the information to readers, that check could perhaps be fine-tuned to: skip entries in the "Editors" list that have a non-empty `<span class="former">` next to the editor's name. This would allow things like `Edith Orre <span class="former">(until September 2025)</span>`, but would prevent `<span class="former">Edith Orre</span>` as it does not convey any information to readers.

deniak:
Fix #2003

### DIFF
--- a/test/data/TR/TRBase.js
+++ b/test/data/TR/TRBase.js
@@ -40,6 +40,14 @@ export const rules = {
                 data: 'noEditorParticipation',
                 errors: ['headers.editor-participation.not-participating'],
             },
+            {
+                data: 'formerEditorParticipation',
+                errors: [],
+            },
+            {
+                data: 'formerEditorParticipationNoRationale',
+                errors: ['headers.editor-participation.not-participating'],
+            },
         ],
     },
     style: {
--- a/test/doc-views/layout/spec.handlebars
+++ b/test/doc-views/layout/spec.handlebars
@@ -102,7 +102,7 @@
                         <dd class="p-author h-card vcard" data-editor-id="{{dl.editor.id}}">George Herald (WebFoo)
                         </dd>
                         {{#if dl.editor2.show}}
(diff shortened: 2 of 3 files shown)


deniak:
@tidoust I've started implementing the patch to allow former editors to be listed and I'm forcing the `<span class="former">` to start with the text '(until ' to avoid having these editors being listed just like the others.
That said, I can imagine some groups would want to provide different kind of reasons, e.g. https://www.w3.org/TR/mathml4/.
So I'm wondering if forcing that specific text is acceptable. It will still be an improvement compared to what we have today.

tidoust:
Yes, I would personally only check that the span is not empty (after trimming) for the time being, to give groups enough leeway. It wouldn't check that the information is meaningful but we've been publishing specs for years without even looking into this, and I don't think anyone has willingly abused the system until now. It seems fine to trust people. As staff contact, what I really want is to trap the case when an active editor has not yet joined the working group.

PR #2085, opened 2026-04-07
Reinstate successful validate API tests

kfranqueiro:
This restores 2 API unit test cases that were commented out several years ago, using one of the generated "good" tests produced by our test server.

I also renamed the `metadata` subdirectory under `test/docs` to `api` since it was presently a misnomer (the only document currently under it is used to test both `/api/metadata` and `/api/validate`), and stored a document for the new test case there. (I couldn't reuse one of the files under `test/docs/` because now that they've been hollowed out, they are usable only for metadata, not validation.)

Getting these cases to work within the same suite required a couple more `persist()` calls among the mocked data setup that happened to not be necessary before.


deniak (commented):
The PR looks good to me. However, I think we are testing a use case that should never happen, e.g. passing a `file` to the validate endpoint. That endpoint should only accept `url`.
Of course, we want to keep the [`validate` function](https://github.com/w3c/specberus?tab=readme-ov-file#validateoptions) processing URLs, files and sources but it would be good to rework these tests and update the [`validate`](https://github.com/w3c/specberus/blob/main/app.js#L107) and [`extractMetadata`](https://github.com/w3c/specberus/blob/main/app.js#L60) events to only accept URLs.

kfranqueiro:
> The PR looks good to me. However, I think we are testing a use case that should never happen, e.g. passing a `file` to the validate endpoint. That endpoint should only accept `url`. Of course, we want to keep the [`validate` function](https://github.com/w3c/specberus?tab=readme-ov-file#validateoptions) processing URLs, files and sources but it would be good to rework these tests and update the [`validate`](https://github.com/w3c/specberus/blob/main/app.js#L107) and [`extractMetadata`](https://github.com/w3c/specberus/blob/main/app.js#L60) events to only accept URLs.

Good point, though this was also true of two tests that have already been running regularly.

I have replaced all of the tests that used the `file` GET parameter with file upload tests. This took longer than I would have liked, largely because I assumed I was using `superagent` incorrectly when the actual problem was that the server that the API tests set up doesn't include the express-fileupload middleware... so I've added some notes to that effect as well.

deniak (approved):
Great, thanks for updating the tests!

PR #2089, opened 2026-04-09
README: Update cjs→mjs syntax; correct Specberus API details

kfranqueiro:
I noticed some parts of the README that seemed out of date.

- Updates CommonJS syntax to ESM syntax to match the codebase
- Removes mention of running Specberus under other environments - I'm not sure how old that note is, but the codebase is pretty hard-committed to Node.js at this point
- Updates details about the Specberus API, since what was written in that section didn't seem to match what's actually exposed anymore; this should be further improved upon migrating to TypeScript when we can emit typings

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,17 +158,16 @@ $ RULE=copyright TYPE=noCopyright npm run test
 
 ## 4. JS API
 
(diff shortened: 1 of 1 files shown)


deniak (approved):
Thanks for the update

PR #2090, opened 2026-04-10
Migrate files under lib to TypeScript

kfranqueiro:
This PR contains the first phase of migrating this repository to TypeScript (v6). Tests will be migrated separately.

(EDIT: After initial approval, we will probably keep this open or target a separate branch and PR the subsequent phases against it, e.g. test conversion and more consistent usage of promises, in order to consolidate all of the related changes into a single major version bump.)

Note that this presently includes a port of #2085; I will rebase once that is merged so it should no longer show up in the diff.

GitHub will display some JS→TS conversions as full file replacements; see [this diff](https://diff2html.xyz/demo?matching=none&matchWordsThreshold=0.25&maxLineLengthHighlight=10000&diffStyle=word&colorScheme=dark&renderNothingWhenEmpty=0&matchingMaxComparisons=2500&maxLineSizeInBlockForComparison=200&outputFormat=side-by-side&drawFileList=true&synchronisedScroll=1&highlight=1&fileListToggle=1&fileListStartVisible=0&highlightLanguages=%5Bobject%20Map%5D&smartSelection=1&fileContentToggle=1&stickyFileHeaders=1&diff=https://gist.githubusercontent.com/kfranqueiro/c49c1d01de3c2c9d68f726a8a7f94aaf/raw/5470ae6e8ac8327d8c7ee9cbe166758ba040b447/specberus.diff) for a better view (generated with `git diff -w -M2 main`). Note that some diffs still get weird here (e.g. `lib/rules/headers/copyright.ts`), and it can't expand diff-less sections (since they're literally not in the diff file).

Unfortunately, `validator.ts` is still marked as a full replacement even in the reduced diff, probably because there's enough dislocation due to the addition of typings and conversion to a class, so that will probably need to be viewed side-by-side with the [current version](https://github.com/w3c/specberus/blob/main/lib/validator.js) manually. It still defines the same functions in the same order.

## Notable Changes

- `validator.ts` now defines `Specberus` as a class rather than a function + prototype (its consumption is still the same)
- Common types are defined in `lib/types.d.ts` (named with `.d.ts` so it doesn't emit an empty module)
- All rules' `check` functions are exported typed as `RuleCheckFunction`, which automatically provides typings for its parameters
  - In instances where `done` is passed a parameter, the type of that parameter is specified via generic
- Fully removes `validator#loadDocument` and the `document` property/param, which seemed entirely unused (previously encountered in #2038)
- Removes `sink.js`; all instances now use `EventEmitter` directly
- Removes the `tools` folder, whose scripts seemed unused for several years
  - Also removes `lib/groups-db.json` which was output by one of these tools but never referenced anywhere else
- Refactors some of `lib/api.ts` to reduce repetition and unnecessary function parameters
- Refactors `lib/exceptions.ts` to no longer export a constructor, since all it entails is a single static function
- Replaces all instances of `Object.prototype.hasOwnProperty.call` with `Object.hasOwn`
- Replaces `importJSON` utility function with `import data from "./file.json" with { type: "json" }`
- Adds/updates various npm scripts; `npm run live` continues to work (now using `tsx` rather than `nodemon`)
- Adds `files` to `package.json` so that npm publishes will include only the built `.js` and `.d.ts` files, and will also no longer include `test` files, dotfiles, etc.
  - Note this means `npm run build` should be run _before_ publishes
- Removes eslint, which was effectively unused due to its configuration not being exported, similar to what happened in spec-generator
- Updates `tsconfig` for stricter checks (akin to what was done in spec-generator) and TS 6; all code passes TS strict mode (which v6 now makes the default)

One thing you might see sprinkled throughout is the use of `!`, especially on e.g. `sr.config!`. This is necessary because it refers to a property that is not initialized within the constructor so it is not guaranteed to be defined, but we expect it to be defined before rule functions are invoked. I tried to minimize how much this was necessary; avoiding it outright would most likely require a deeper rethink of the exposed API (or at minimum the API that rule functions are built around). While this PR includes some refactors, partly to make typings work more smoothly, it aims to avoid major rewrites.


package.json
```diff
@@ -9,61 +9,67 @@
...
+    "compression": "1.8.1",
+    "cors": "2.8.6",
     "cspell": "9.0.2",
     "domhandler": "^6.0.1",
-    "eslint": "10.0.3",
-    "eslint-config-prettier": "10.0.3",
-    "eslint-plugin-node": "11.1.0",
-    "eslint-plugin-prettier": "5.0.1",
     "expect.js": "0.3",
+    "express": "5.2.1",
```
deniak: most of the dependencies you moved under `devDependencies` are still needed in production.

deniak (changes requested):

package.json
kfranqueiro: After further discussion, I've removed the last 3 commits that involved this change and the others related to omitting `app.js` from publishes, since we rely on `npm i --production` for the server deploy (which does need `app.js` and its dependencies).

lib/api.ts
```diff
@@ -0,0 +1,172 @@
...
+    info: HandlerMessage[] = [],
+    metadata: Record<string, string> = {}
+) {
+    delete metadata.file;
+    const wrapper = buildJSONresult(errors, warnings, info, metadata);
+    res.status(wrapper.success ? 200 : 400).json(wrapper);
+};
+
+const getFullUrl = (req: Request) =>
+    new URL(`http://${process.env.HOST ?? 'localhost'}${req.url}`);
```
deniak: We don't rely on the `HOST` environment variable. It might be safer to just use `req.host`

```suggestion
    new URL(`http://${req.host ?? 'localhost'}${req.url}`);
```

deniak (changes requested):

lib/api.ts
kfranqueiro: Thanks for spotting this; I've improved it to use `req.host` and also `req.protocol`. (We only use the pathname portion so it would've gone unnoticed for a while.)

deniak (approved):
Nice job!

kfranqueiro:
I pushed one more commit to address code which was assuming a direct mapping between *.js files and modules, which no longer holds up when running through tsx (as is done for `npm run live` in this PR, and will be done for `npm test` in the subsequent PR to migrate the tests to TypeScript).

I've verified that `npm run live` now works correctly even without a previous build having been run, and have also verified that all of the tests on my upcoming branch pass successfully with this commit in place.

lib/util.ts
```diff
@@ -41,32 +36,48 @@ export const buildJSONresult = function (
...
     }
 }
 
 export const allProfiles = [
     ...TRProfiles.map(x => `TR/${x}`),
-    ...filesSUBM.map(x => `SUBM/${x}`),
+    ...SUBMProfiles.map(x => `SUBM/${x}`),
 ];
 
 export const profiles = Object.fromEntries(
```
deniak: This doesn't seem to be used anywhere.

deniak (changes requested):

lib/util.ts
kfranqueiro: Good catch; I've now removed it. Perhaps in the past this was used to reference profile modules, but now we seem to go through `allProfiles` and import where we use them.

deniak (approved):

lib/util.ts
deniak: addressed in https://github.com/w3c/specberus/pull/2090/commits/bb20f1ddbcc2f49fc6be37bf62b9f10d42ba0e2d

PR #2097, opened 2026-05-06
Convert JSON files to TypeScript modules

kfranqueiro:
This converts several JSON files in the repo to TypeScript modules, allowing us to define (and thus hold ourselves to) typings for them, and in some cases improve how they are defined/formatted.

I suggest viewing the diff with whitespace ignored. The migrated files will still show up as full rewrites due to no longer being JSON (and thus keys no longer being quoted), but none of the data has changed. All tests still pass with no substantive changes.

Note I have not done any further reformatting of any strings in these files, so it should still be somewhat feasible to compare the old and new versions side-by-side manually.

Additionally:

- `badterms.json` is migrated from `public` to `lib` as it is also used by a rule; it is still exposed from the pubrules server at `/badterms.json`
- `rules.json` is split into two files to simplify code that consumes it:
  - `rules-generic-sections.ts` for the data formerly under the `*` key
  - `rules-track.ts` for everything else falling under a particular track
- `exceptions.json` is converted to a `Map` so that the keys can be directly expressed as regular expressions, rather than needing to instantiate a new `RegExp` every time
- A few existing types are moved into the new `ts` files to live closest to their related data
- I exported a centralized `specberusVersion` from `lib/util.ts` so that various modules don't all have to look it up from `package.json`
- I discovered that the recursive call in `lib/exceptions.ts` wasn't being used at all because the condition would never be met, which enabled further simplifying that module


app.ts
```diff
@@ -41,7 +39,6 @@ const io = new Server(server);
 // Middleware:
 app.use(morgan('combined'));
 app.use(compression());
-app.use('/badterms.json', cors());
```
deniak, line 44: `public/badterms.json` can now be deleted

deniak (changes requested):

app.ts
kfranqueiro: Whoops thanks, can't believe I somehow missed deleting the original file; fixed.

deniak (approved):

PR #2098, opened 2026-05-06
Convert tests to TypeScript and remove Node v20 support

kfranqueiro:
This follows up #2090 by converting the `tests` folder to TypeScript. At this point, `npm test` and `npm run testserver` run through `tsx` and no longer require a build.

I am once again providing a [separate diff view](https://diff2html.xyz/demo?matching=none&matchWordsThreshold=0.25&maxLineLengthHighlight=10000&diffStyle=word&colorScheme=dark&renderNothingWhenEmpty=0&matchingMaxComparisons=2500&maxLineSizeInBlockForComparison=200&outputFormat=side-by-side&drawFileList=true&synchronisedScroll=1&highlight=1&fileListToggle=1&fileListStartVisible=0&highlightLanguages=%5Bobject%20Map%5D&smartSelection=1&fileContentToggle=1&stickyFileHeaders=1&diff=https://gist.githubusercontent.com/kfranqueiro/e4563b2b07e89b740b0d0a5c133386a6/raw/2b8b498413bf9112ccc45412c4133444a4ff3197/specberus-tests.diff) based on a local `git diff -w -M2` to avoid misflagging moves as rewrites.

## Notable changes

- Fixes/improves a few issues in the tests:
  - The `type=good2` tests were actually duplicating the `type=good` tests, which I stumbled across when simplifying loops in the tests. Moreover, there were several mistakes that meant these would never have passed. The data and expectations involved are now corrected so that these render correctly and pass.
  - A couple of mistakes in inherited data were flagged by TypeScript and have been corrected; I have to assume they were inconsequential to existing tests...
  - The `buildBadTestCases` flow in `test/lib/utils.ts` has been made a bit more digestible in an effort to make its typings flow together naturally; in particular, `buildTrackTestCases` now does the one thing it says it does, and `buildDocTypeTestCases` no longer calls it in cases where it shouldn't
- Added a few new meaningful assertions, particularly in tests that do nothing besides assert something is an object (which seems extremely unlikely to ever fail).
- Removes `test/validation.js` which seemed to be unused and referenced files that do not exist
- Removes nodemon, mocha, and chai, instead relying on node's own assert and test APIs (reduces `node_modules` by 8MB)
- Removes Node v20 support, an idea I previously floated past Denis. Security support for v20 just ended at the end of April, and supporting it while making use of `node:test` would require extra effort that doesn't seem warranted.
  - This PR does go out of its way to support Node v22 in the `coverage` script used by CI:
    - Avoids passing `--test-isolation=none` which is only supported by v24 (more on this below)
    - Passes `--test-timeout` with a 2-minute timeout, as this value seems to affect the entire suite in Node v22, whereas in v24 I get the impression it only affects each individual test (though I was not able to find documentation of this changing)
  - RE `--test-isolation=none`: `node:test` now defaults to running every test in a separate child process. We were not relying on this before when using mocha, so I see no need to rely on it now. When process isolation is used, `describe.only` and `it.only` require an extra CLI flag, `--test-only`, in order to work; this is not necessary when process isolation is disabled.


deniak:
>   * This PR does go out of its way to support Node v22 in the `coverage` script used by CI:
>     [...]
>     * Passes `--test-timeout` with a 2-minute timeout, as this value seems to affect the entire suite in Node v22, whereas in v24 I get the impression it only affects each individual test (though I was not able to find documentation of this changing)

The bug was reported [here](https://github.com/nodejs/node/issues/57656). Unfortunately, the [fix](https://github.com/nodejs/node/pull/57672) only landed in [node 24](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V24.md#user-content-24.0.0) so I'm afraid we need to keep that long timeout until we drop support for node 22 😞

deniak (commented):
The PR looks good to me. I just noticed something that is not related to your changes but we still have 3 unmocked requests with the API tests. Any chance you can add these?
```
▶ API
  ▶ Endpoint
    ✖ Should exist and listen to GET requests (22.423833ms)
    ✔ Should exist and listen to POST requests (5.597125ms)
  ✖ Endpoint (28.4455ms)
  ▶ Method “version”
    ✔ Should return the right version string (2.145167ms)
  ✔ Method “version” (2.23375ms)
  ▶ Method “metadata”
    ✔ Should accept "file" via POST, and return the right profile and date (36.871042ms)
  ✔ Method “metadata” (36.982833ms)
Unmocked non-local request: https://validator.w3.org/nu/?out=json ReadableStream { locked: false, state: 'readable', supportsBYOB: false }
  ▶ Method “validate”
    ✔ Should 400 and return an array of errors when validation fails (1259.598083ms)
Unmocked non-local request: https://validator.w3.org/nu/?out=json ReadableStream { locked: false, state: 'readable', supportsBYOB: false }
    ✔ Should accept "file" via POST, and succeed when the document is valid (428.01875ms)
Unmocked non-local request: https://validator.w3.org/nu/?out=json ReadableStream { locked: false, state: 'readable', supportsBYOB: false }
    ✔ Special profile “auto”: should detect the right profile and validate the document (426.79625ms)
```

kfranqueiro:
> The PR looks good to me. I just noticed something that is not related to your changes but we still have 3 unmocked requests with the API tests. Any chance you can add these?

This isn't an accident; I specifically noted in #2048 that I didn't mock the HTML checker service request(s). I can work on that, but I'd like to do it separately from this PR. I'll need to think about the best way to mock them, since the requests are only distinguishable by their POST data, and currently we're setting up mocks for the API tests once at the entire suite level.

Edit: Added a mock in #2102

deniak (approved):

PR #2099, opened 2026-05-06
Remove Node.js version 20 from CI workflow

deniak:
We are going to remove support for node 20 in an upcoming release.
That PR updates the CI to only run with node 22 and 24 to avoid blocking the PR for that future release.


deniak:
Closing in favor of #2100

PR #2106, opened 2026-05-19
Convert validate, extractMetadata, and rule check APIs to return promises

kfranqueiro:
**This PR contains a significant amount of API changes, some of them breaking,** but the result should be easier to use and avoids hanging due to unexpected errors.

- Converts `extractMetadata`, `validate`, and all rule check functions to return promises
  - This involves re-flowing code in a few rules to return existing promises they already produced
  - Rules no longer use `sr.throw`; they throw errors directly, and this is wrapped with `this.#throw` on the `validator.ts` end (the `throw` method is now private)
    - Any `#throw` invocations are collected into a list of exceptions, which will be included within the rejected promise's error in an `exceptions` property
  - Rules which checked for superagent timeouts now compare correctly (as alerted by superagent typings)
- Re-flows asynchronous `validator.ts` APIs that cache values to cache the pending promise, to avoid duplicated near-simultaneous requests
- Any `validator.ts` APIs that were not intended to be directly used are now private
- Removes `meta` instance property (use the resolved value of `extractMetata` instead)
- Removes `clearCache` method which was never used (all usage only calls `extractMetadata` or `validate` once)
- Removes unused optional parameter from `extractHeaders` method
- Removes `start-all` and `end-all` events (use promises instead for `end-all`; `start-all` already seemed unused)
- `extractMetadata` and `validate` now emit error/info/warning events consistently with each other (previously, `extractMetadata` did not include `key`, `extra`, or `detailMessage`)
- Removes `events` property from `extractMetadata` and `validate` options; use `sr.on(...)` (which exposes specific typings for supported events) instead of passing an entire `EventEmitter`
- `exception` handler is no longer required in Specberus options; exceptions will reject the promise, which should be handled instead
  - `exception` is still a listenable event for instances where granular feedback is desired
- When exceptions are reported via errors in API output, they are now objects with 'exception' key, to be more consistent with all other reported errors being objects (not strings)
- Removes `doasync` dependency (used by shortname rule) which was not doing anything necessary (`superagent` methods already return promise-likes)
- Converts all `fs` usage in `validator.ts` to use `fs/promises`
- Reduces duplicated code within `validator.ts` with new private functions such as `#load`, `#prepare`, and `#reportResult`
- Adds a few test cases (some long overdue)
- Restores and fixes `type=good2` test cases which were not actually being run (instead they were re-running `type=good` cases)

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -7,7 +7,6 @@ import compression from 'compression';
 import cors from 'cors';
 import express from 'express';
 import fileUpload from 'express-fileupload';
-import EventEmitter from 'events';
 import { writeFile } from 'fs';
 import http from 'http';
 // @ts-ignore (no typings)
@@ -67,8 +66,7 @@ io.on('connection', socket => {
                 message: 'URL or file not provided.',
             });
         const specberus = new Specberus();
-        const handler = new EventEmitter();
-        handler.on('err', (type, data) => {
+        specberus.on('err', (type, data) => {
             try {
                 socket.emit(
                     'err',
@@ -78,7 +76,7 @@ io.on('connection', socket => {
                 socket.emit('exception', err.message);
             }
         });
-        handler.on('warning', (type, data) => {
+        specberus.on('warning', (type, data) => {
             try {
                 socket.emit(
                     'warning',
@@ -88,7 +86,7 @@ io.on('connection', socket => {
                 socket.emit('exception', err.message);
             }
         });
-        handler.on('info', (type, data) => {
+        specberus.on('info', (type, data) => {
             try {
                 socket.emit(
                     'info',
@@ -98,10 +96,9 @@ io.on('connection', socket => {
                 socket.emit('exception', err.message);
             }
         });
-        handler.on('exception', data => {
+        specberus.on('exception', data => {
             socket.emit('exception', data);
         });
-        data.events = handler;
         try {
             const metadata = await specberus.extractMetadata(data);
             socket.emit('finishedExtraction', {
@@ -135,12 +132,11 @@ io.on('connection', socket => {
             });
         }
         const specberus = new Specberus();
-        const handler = new EventEmitter();
         const profileCode = profile.name;
         socket.emit('start', {
             rules: (profile.rules || []).map(rule => rule.name),
         });
-        handler.on('err', (type, data) => {
+        specberus.on('err', (type, data) => {
             try {
                 socket.emit(
                     'err',
@@ -150,7 +146,7 @@ io.on('connection', socket => {
                 socket.emit('exception', err.message);
             }
         });
-        handler.on('warning', (type, data) => {
+        specberus.on('warning', (type, data) => {
             try {
                 socket.emit(
                     'warning',
@@ -160,7 +156,7 @@ io.on('connection', socket => {
                 socket.emit('exception', err.message);
             }
         });
-        handler.on('info', (type, data) => {
+        specberus.on('info', (type, data) => {
             try {
                 socket.emit(
                     'info',
@@ -170,10 +166,10 @@ io.on('connection', socket => {
                 socket.emit('exception', err.message);
             }
         });
-        handler.on('done', name => {
+        specberus.on('done', name => {
             socket.emit('done', { name });
         });
-        handler.on('exception', data => {
+        specberus.on('exception', data => {
             socket.emit('exception', data);
         });
 
@@ -192,7 +188,6 @@ io.on('connection', socket => {
                             await specberus.validate({
                                 url: data.url,
                                 profile,
-                                events: handler,
                                 validation: data.validation,
                             });
                         } catch (e) {
@@ -220,7 +215,6 @@ io.on('connection', socket => {
                 await specberus.validate({
                     file: data.file,
                     profile,
-                    events: handler,
                     validation: data.validation,
                 });
             } finally {
--- a/app.ts
+++ b/app.ts
@@ -12,14 +12,14 @@ import http from 'http';
 // @ts-ignore (no typings)
 import insafe from 'insafe';
 import morgan from 'morgan';
-import { Server } from 'socket.io';
+import { Server, type Socket } from 'socket.io';
 import tmp from 'tmp';
 
 import * as api from './lib/api.js';
 import badterms from './lib/badterms.js';
 import * as l10n from './lib/l10n.js';
 import { allProfiles, specberusVersion } from './lib/util.js';
-import { Specberus } from './lib/validator.js';
+import { ExceptionsError, Specberus } from './lib/validator.js';
 import * as views from './lib/views.js';
 import type { ProfileModule } from './lib/types.js';
 
@@ -57,6 +57,18 @@ views.setUp(app);
 l10n.setLanguage('en_GB');
 
 server.listen(process.argv[2] || process.env.PORT || DEFAULT_PORT);
+
+/** Emits misc. errors, for cases where 'exception' handler already emits individual exceptions. */
+function reportNonExceptionsError(
+    socket: Socket,
+    e: any,
+    subject: 'Metadata extraction' | 'Validation'
+) {
+    if (!(e instanceof ExceptionsError))
+        socket.emit('exception', {
+            message: `${subject} encountered an unexpected error: ${e}`,
+        });
+}
 
 io.on('connection', socket => {
     socket.emit('handshake', { version: specberusVersion });
@@ -105,8 +117,8 @@ io.on('connection', socket => {
                 ...metadata,
                 url: data.url,
             });
-        } catch {
-            // exceptions already emitted from event handler
+        } catch (e) {
+            reportNonExceptionsError(socket, e, 'Metadata extraction');
         }
     });
     socket.on('validate', async data => {
@@ -191,9 +203,7 @@ io.on('connection', socket => {
                                 validation: data.validation,
                             });
                         } catch (e) {
-                            socket.emit('exception', {
-                                message: `Validation blew up: ${e}`,
-                            });
+                            reportNonExceptionsError(socket, e, 'Validation');
                         } finally {
                             socket.emit('finished');
                         }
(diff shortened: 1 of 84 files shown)


lib/validator.ts
```diff
@@ -387,30 +364,30 @@ export class Specberus {
...
-        if (this.$docDateEl) return this.$docDateEl;
+        if (this.#$docDateEl) return this.#$docDateEl;
         this.getDocumentDate();
-        return this.$docDateEl;
+        return this.#$docDateEl;
     }
 
     getSotDSection() {
-        if (typeof this.$sotdSection !== 'undefined') return this.$sotdSection;
+        if (this.#$sotdSection) return this.#$sotdSection;
```
deniak: The cache will not work as expected if there's no SotD found, i.e. `this.#sotdSection === null`.

```suggestion
        if (this.#$sotdSection !== undefined) return this.#$sotdSection;
```

The same comment also applies for a few functions below such as `getChartersData`, `getDelivererIDs`, etc

lib/validator.ts
```diff
@@ -849,89 +774,81 @@ export class Specberus {
...
-        fs.access(file, fs.constants.F_OK, errors => {
-            if (errors) return cb(`File '${file}' not found.`);
-            fs.readFile(file, { encoding: 'utf8' }, (err, src) => {
-                if (err) return cb(err);
-                this.loadSource(src, cb);
-            });
-        });
+    async #loadFile(file: string) {
+        try {
+            access(file, constants.F_OK);
```
deniak: Now that we switched to `fs/promises`, I think we should wait for it:
```suggestion
            await access(file, constants.F_OK);
```

lib/api.ts
```diff
@@ -88,21 +111,24 @@ const processPost = () => async (req: Request, res: Response) => {
...
+        result => {
+            sendResult(
+                res,
+                metadataOverride
+                    ? { ...result, metadata: metadataOverride }
+                    : result
+            );
+        },
+        (error: ExceptionsError) => {
+            sendErrors(res, error.exceptions, 'exception');
```
deniak: IIUC, we may have cases where `error.exceptions` could be undefined.
The `validate` and `extractMetadata` promises may throw simple `Error`. For instance,

```js
    async validate(options: ValidateOptions) {
        if (!options.profile)
            throw new Error('Without a profile there is nothing to check.');
```

or

```js
    #load(options: ExtractMetadataOptions | ValidateOptions) {
        if (options.url) return this.#loadURL(options.url);
        if (options.source) return this.#loadSource(options.source);
        if (options.file) return this.#loadFile(options.file);
        throw new Error('url, source, or file must be specified.');
    }
```

In these cases, I believe `sendErrors` would fail.

app.ts
```diff
@@ -185,19 +182,19 @@ io.on('connection', socket => {
...
                         try {
-                            specberus.validate({
+                            await specberus.validate({
                                 url: data.url,
                                 profile,
-                                events: handler,
                                 validation: data.validation,
                             });
                         } catch (e) {
                             socket.emit('exception', {
```
deniak: I suppose this is redundant now that there's the 'exception' listener l. 172

lib/validator.ts
```diff
@@ -15,30 +15,22 @@ import { assembleData, setLanguage } from './l10n.js';
...
     RuleMeta,
     SpecberusConfig,
-    ProfileModule,
 } from './types.js';
 
 setLanguage('en_GB');
 
 interface BaseOptions {
-    events: EventEmitter;
+    events?: EventEmitter;
```
deniak: Are you keeping this for backward compatibility? If that's the case and we expect to deploy a new major version with breaking changes, I think we can remove this.

deniak (changes requested):

lib/api.ts
kfranqueiro: I had gone down a bit of a rabbit hole regarding this same realization, and I think most of the types of error that would cause this would already be prevented by other checks done by `api.ts` before `sendErrors` gets called (e.g. because `api.ts` already handles calling `processParams` directly first), but I did want to double-check and test a few things, so thanks for this call-out.

lib/validator.ts
kfranqueiro: I've done another pass over these across the file, and I think this is the only removed check that needs to be restored; the others that have been removed might return empty arrays (or promises resolving to them), but never null or anything else falsy. I must have gotten carried away with this one while I was cleaning up the others.

lib/validator.ts
kfranqueiro: Yup, I had intended to do this and then somehow neglected to add it. Thanks for catching.

lib/validator.ts
kfranqueiro: Nope, it's entirely unused; I had initially set it to optional as I was starting to remove it, then forgot to completely remove it once I finished. Thanks for catching.

app.ts
kfranqueiro: Wouldn't this be necessary for the same reason as your remark above, in that `validate` could throw an error before it even gets to rule checking?

lib/api.ts
kfranqueiro: OK, I went down far less of a rabbit hole today on this subject, and better centralized logic within `sendErrors` to handle both generic error and ExceptionsError cases. This was feasible since the previously-separate flows for each type of case behaved consistently.

I also added a trivial unit test that I'm surprised we didn't already have for a very straightforward non-exception error path, though I will point out this particular test case already passed before because it went down one of the existing `error.toString()` branches.

lib/validator.ts
deniak: thanks. Addressed in https://github.com/w3c/specberus/pull/2106/changes/a2c6107c2ba9990efc45faa71d11dcbb54320365

lib/validator.ts
deniak: addressed in https://github.com/w3c/specberus/pull/2106/changes/e3d10fd640b809a13736d6285b49ff39be7cc4a0

lib/api.ts
```diff
@@ -2,35 +2,51 @@
...
-    warnings: HandlerMessage[] = [],
-    info: HandlerMessage[] = [],
-    metadata: Record<string, string> = {}
-) {
-    delete metadata.file;
-    const wrapper = buildJSONresult(errors, warnings, info, metadata);
-    res.status(wrapper.success ? 200 : 400).json(wrapper);
-};
+function sendErrors(res: Response, error: Error | ExceptionsError) {
+    res.status(error instanceof ExceptionsError ? 400 : 500).json({
```
deniak: That's inconsistent with the jsdoc. I guess you mean

```suggestion
    res.status(error instanceof ExceptionsError ? 500 : 400).json({
```

lib/validator.ts
```diff
@@ -117,214 +108,184 @@ const abbrMonths = [
...
-                '[WARNING] No handler for event `exception` which to report system errors.'
-            );
+        await Promise.all(
+            profile.rules.map(async rule => {
+                try {
+                    const result = await rule.check(this);
+                    if (result)
+                        for (const [key, value] of Object.entries(result))
+                            metadata[key] = value;
+                    this.emit('done', rule.name);
```
deniak: I think it's better to move that line in a `finally` block so when we run the check from the UI, the progress bar isn't stuck if a rule throws.
The same comment in the `validate` function.

lib/util.ts
deniak: I guess this can be removed since `events` is no longer an option.

deniak (changes requested):

lib/validator.ts
kfranqueiro: Discussed and agreed upon during May 27 call.

* Behavior (and test) updated in [3a119a8](https://github.com/w3c/specberus/pull/2106/commits/3a119a80c3975da1e237ca8a7daa3d6d0d9a3fe1)
* Clarification included in latest README pass in [7b22185](https://github.com/w3c/specberus/pull/2106/commits/7b221859511bb9d83cd74255576d41d5bfe5adce)

FYI, I did some more testing on the current observable UI behavior (prior to the above commits):

- The progress bar is only shown for validation runs, not for the up-front `extractMetadata` run when profile is set to auto
- The only explicitly-coded exceptions that will ever be reported during validation (not metadata) relate to HTML validation
- Even if an exception occurs, it does not impact the displayed number of checks run in the final output, which is directly based on the length of the profile's `rules` array (tested by intentionally throwing an error in a traversed code path)

lib/api.ts
kfranqueiro: Welp, that's an embarrassing goof-up, which I probably fell for due to the ordering of the previous line vs. this one. I've matched up the ordering, fixed this, and added verification of HTTP status code to the API tests in b4f96f2.

lib/util.ts
kfranqueiro: I've removed this and other values in this condition that appeared to be completely unused in the repo, in https://github.com/w3c/specberus/pull/2106/commits/0b8ab287fc194903352f28b14ba9b07421f4c486.

app.ts
deniak: For the rules, there's a `try/catch` block in `validator.ts` l.243:
```ts
                try {
                    const result = await rule.check(this);
                    if (result)
                        for (const [key, value] of Object.entries(result))
                            metadata[key] = value;
                } catch (error) {
                    this.#throw(error.message);
                } finally {
                    this.emit('done', rule.name);
                }
```

and the `throw` function will emit the exception so we should be safe.

However, while reviewing the different places where an exception is thrown, I think the `validate` function may throw without emitting the exception (see `validator.ts` l.260).

To avoid the duplication, how about checking the type of the exception?
```ts
} catch (e) {
    if (!(e instanceof ExceptionsError)) {
        // ExceptionsError: exceptions already forwarded individually via the event listener
        // Error: not emitted yet (e.g. missing profile), so emit it now
        socket.emit('exception', { message: `Validation blew up: ${e}` });
    }
} finally {
    socket.emit('finished');
}
```

kfranqueiro: Good point RE duplication. I also did a pass over each place we're calling `extractMetadata` and `validate` and made all of them consistently report non-`exception` errors, in c45e487.

deniak (approved):
Great job!

PR #2114, opened 2026-05-27
Split main Specberus and rule check APIs into separate classes

kfranqueiro:
This separates the APIs responsible for creating/running metadata extraction or validation from the APIs necessary to perform the various rule checks. This results in a far simpler API for typical consumers who only care about the former, and prevents each use case from crossing wires with one another.

## Changes

- Separates APIs used by rules to a `RuleContext` class, which now lives in `lib/rule-context.ts`
- Moves the remainder of the code in `validator.ts` (for the `Specberus` class) to `specberus.ts`, and updates `main` in `package.json` to match
- Updates all usage of "sr" as a variable name within rule check functions to "context"
- Updates nomenclature in the readme; also replaces outdated future plans sentence

The diff will appear larger than it really is due to moving existing code from `validator.ts` to `specberus.ts` and `rule-context.ts`, but no functional changes have been made to the code that previously lived in `validator.ts`.

### DIFF
--- a/lib/rule-context.ts
+++ b/lib/rule-context.ts
@@ -9,6 +9,7 @@ import w3cApi from 'node-w3capi';
 
 import { hasExceptions } from './exceptions.js';
 import { assembleData } from './l10n.js';
+// Reminder: specberus.js imports this module; only types may be imported in the other direction
 import type { Specberus } from './specberus.js';
 import { get } from './throttled-ua.js';
 import { AB, REC_TEXT, specberusVersion, TAG } from './util.js';
--- a/lib/rule-context.ts
+++ b/lib/rule-context.ts
@@ -2,241 +2,24 @@
  * @file Main file of the Specberus npm package.
  */
 
-import EventEmitter from 'events';
-import { access, constants, readFile } from 'fs/promises';
-
 import { type Cheerio, type CheerioAPI, load } from 'cheerio';
 import type { Element } from 'domhandler';
 // @ts-ignore (no typings)
 import w3cApi from 'node-w3capi';
 
 import { hasExceptions } from './exceptions.js';
-import { assembleData, setLanguage } from './l10n.js';
-import * as profileMetadata from './profiles/metadata.js';
-import * as profileAdditionalMetadata from './profiles/additionalMetadata.js';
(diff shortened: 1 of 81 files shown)


lib/types.d.ts
```diff
@@ -64,7 +65,7 @@ export interface RecMetadata {
     rectrack?: string | null;
 }
 
-export type RuleCheckFunction<R = void> = (sr: Specberus) => R | Promise<R>;
+export type RuleCheckFunction<R = void> = (sr: RuleContext) => R | Promise<R>;
```
deniak:
```suggestion
export type RuleCheckFunction<R = void> = (context: RuleContext) => R | Promise<R>;
```

lib/rule-context.ts
```diff
@@ -2,48 +2,25 @@
...
 import type { Element } from 'domhandler';
 // @ts-ignore (no typings)
 import w3cApi from 'node-w3capi';
 
 import { hasExceptions } from './exceptions.js';
-import { assembleData, setLanguage } from './l10n.js';
-import * as profileMetadata from './profiles/metadata.js';
-import * as profileAdditionalMetadata from './profiles/additionalMetadata.js';
+import { assembleData } from './l10n.js';
+import type { Specberus } from './specberus.js';
```
deniak, line 13: I suppose the `import type` here is to avoid a circular `import` since `specberus.js` also has `import { RuleContext } from './rule-context.js';`
If that's the case, it would be good to add a comment explaining this.

deniak (approved):

lib/types.d.ts
kfranqueiro: :man_facepalming: of course I go through all the rules and forget to apply the same change to the typing itself... thanks for catching.

lib/rule-context.ts
kfranqueiro: I can add a comment, though for what it's worth, VSCode/VSCodium automatically knows enough to `import type` in situations where only the type is explicitly referenced and `verbatimModuleSyntax: true` is listed in `tsconfig.json`. While it indeed avoids a circular import in this particular case, it's also useful in general to avoid importing emitted code in the compiled JS (whereas types are not emitted).

PR #2119, opened 2026-06-04
Rules tests: Prevent empty suite noise when filtering via env vars

kfranqueiro:
The way that the filtering checks for the `PROFILE`, `RULE`, and `TYPE` environment variable checks were implemented in `rules.ts` would skip individual tests but not the suites. Mocha would automatically filter out empty suites entirely, but `node:test` does not do this, which results in the few tests that did run being buried between logs for many empty suites.

This PR moves the checks responsible for filtering tests so that entire irrelevant suites are filtered.

Note that this also moves the trimming of `.js` from inside of `runTestsForProfile` to outside of it, to coincide with where filtering is performed.

I verified that `PROFILE`, `RULE`, and `TYPE` still behave as expected (including in combination, and when `PROFILE=MEM-SUBM` is used).


deniak (approved):

deniak:
Good catch!

PR #2167, opened 2026-08-25
Skip tests populated for inapplicable profile/rule combinations

kfranqueiro:
The way test views are populated in this repository causes a few cases to be inherited under profiles that don't actually contain the rules being tested. The tests in question pass anyway, as they are forcibly hooked up to a fabricated profile containing the rule, even though the combination does not reflect reality.

This PR first checks whether the rule being tested actually exists in the profile's real list of rules, and skips any unrealistic combinations.

This results in 18 fewer tests being run; the following combinations are skipped:

- for profiles CR and CR-Echidna: sotd.draft-stability 
- for profile DISC: headers.editor-participation 
- for profile REC-RSCND:
  - sotd.charter 
  - sotd.publish 
  - sotd.stability 
  - sotd.pp 
- for profiles CRY, CRYD, and DRY: sotd.usage

### DIFF
--- a/test/rules.ts
+++ b/test/rules.ts
@@ -321,7 +321,11 @@ function checkRule(tests: RuleTest[], options: CheckRuleOptions) {
         if (testType && test.data !== testType) return;
 
         it(`should ${passOrFail} for ${url}`, async () => {
-            const { config } = resolvedProfiles[profile];
+            const config = resolvedProfiles[profile]?.config;
+            if (!config)
+                throw new Error(
+                    `Config not found for profile ${suffix}; check that the module exists and is valid?`
+                );
             const ruleModule = await import(
                 `../lib/rules/${category}/${rule}.js`
             );
--- a/test/rules.ts
+++ b/test/rules.ts
@@ -5,7 +5,7 @@ import { after, afterEach, before, beforeEach, describe, it } from 'node:test';
 
 import { rules as metadataRules } from '../lib/profiles/metadata.js';
 import { removeRules } from '../lib/profiles/profileUtil.js';
-import type { HandlerMessage } from '../lib/types.js';
+import type { HandlerMessage, ProfileModule } from '../lib/types.js';
 import { allProfiles } from '../lib/util.js';
 import {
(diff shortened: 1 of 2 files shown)


test/rules.ts
```diff
@@ -300,25 +300,28 @@ interface CheckRuleOptions {
...
 }
 
-function checkRule(tests: RuleTest[], options: CheckRuleOptions) {
+async function checkRule(tests: RuleTest[], options: CheckRuleOptions) {
     const { docType, track, profile, category, rule } = options;
 
+    const suffix = track ? `${track}/${profile}` : profile;
+    const { config, rules } = (await import(
+        `../lib/profiles/${docType}/${suffix}.js`
+    )) as ProfileModule;
```
deniak: These imports used to be in a `it` block, which means if it failed for some reasons, the test suite would report an error with the correct exit code `1`. 
Now that the import is outside that block, the error will still be displayed but with an exit code `0` (the test will not be seen as a failed test) which will make the CI report success.

deniak (changes requested):

test/rules.ts
kfranqueiro: Thanks for flagging this; pulling further on this thread led me to ultimately realize that as written at the time, this PR would have been subject to a race condition that it just happened to not be tripping on. I found this out when I tried taking advantage of the `profiles` export in `lib/util.ts`. My choices would have been to either cascade async/await across _all_ involved `describe` calls, or resolve the promises up-front with a single set of top-level `await`s; the latter is much simpler.

To resolve your original concern, I have reinstated the original variables inside the `it`, and put a guard in the up-front skip check so it will fall through if the profile lookup fails, allowing the lookup to then fail inside the `it` as before.

In the process, I also updated the PR to take advantage of the `profiles` export (which IIUC was previously only used by spec-prod), to simplify the instances in `rules.ts` which were previously relying on `allProfiles`. (This was what led me to spot the race condition.)

(I might dig around separately to see if we need `allProfiles` at all, really...)

deniak (approved):

PR #2172, opened 2026-09-14
Add Node 26 to workflow matrix and engines; drop Node 22

kfranqueiro:
This adds `26` to `node-versions` in the Specberus tests workflow, and to `engines` in `package.json`, and (per comment below) drops 22.

Automated tests pass, and I did a bit of testing locally using 26.8.2.

Note that this should probably warrant a major version bump due to dropping a version, especially since 22 is technically not EOL until next year.

Edit: I also edited the lint workflow to stop using a meaningless reference that doesn't exist in that file.


.github/workflows/node.js.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [22, 24]
```
deniak, line 21: I don't think it's worth keeping node 22 in the list. We are already on node 24 and we'll soon upgrade to node 26

nschonni: 22 doesn't hit EOL till 2027-04-30 https://github.com/nodejs/Release/#release-schedule
Less important to support all the Node.js release lines for an application vs. a library though.

deniak (approved):

.github/workflows/node.js.yml
kfranqueiro: I was initially of the same mind as @nschonni RE keeping the EOL in mind. After discussing further with Denis, given that specberus has very few expected dependents, all of which typically run with Node 24 already, I think we are safe to drop Node 22.

That being said, dropping a major Node version (especially before EOL) probably warrants a major version bump. I will make the change to this PR, but am not planning to merge this right away.

For what it's worth, specberus only has 2 known dependents:

- [spec-prod](https://github.com/w3c/spec-prod/), which already requires Node 24 (and [skipped 22 entirely](https://github.com/w3c/spec-prod/pull/210))
- [Echidna](https://github.com/w3c/echidna/), which could stand to have an `engines` update of its own, as it still includes `20`; in practice, we only run it on 24

package.json
```diff
@@ -71,7 +71,7 @@
     "test": "NO_THROTTLE=true tsx --test --test-isolation=none 'test/*.ts'"
   },
   "engines": {
-    "node": "22 || 24",
+    "node": "24 || 26",
     "npm": ">=7"
```
nschonni: Could probably drop this, as I think the minumum `node` version don't support the old NPM version that this was probably trying to work around

package-lock.json
```diff
@@ -48,7 +48,7 @@
         "typescript": "^6.0.3"
       },
       "engines": {
-        "node": "22 || 24",
+        "node": "22 || 24 || 26",
```
nschonni: Doesn't look like the lock file picked up the  same change

kfranqueiro: Guess I must have done that out-of-order, thanks for flagging.

deniak (approved):

